PR #30, opened 2018-01-08
Create the `expansions` table and models using STI

WilliamMcCumstie:
This is a smallish PR that mainly focuses on updating the database to include `expansions`. Each `Expansion` includes an `ExpansionType`, that is stored in it's own table (see: https://github.com/alces-software/alces-flight-center/commit/53903a2976c791fa885363c8f70a54b49cd25173).

Each expansion is either a `ComponentExp..` or `DefaultExp..` expansion. Currently only the associations exist which will expanded on in feature `PR`s. The idea being that a `DefaultExp..` will be used to create a `ComponentExp` when a new `Component` is created (unless otherwise specified).

Another commit is still required to fix `rspec` but further discussion is required.

### DIFF
--- a/app/models/expansion.rb
+++ b/app/models/expansion.rb
@@ -1,3 +1,3 @@
 class Expansion < ApplicationRecord
-  belongs_to :expasion_type
+  belongs_to :expansion_type
 end
--- a/app/models/expansion.rb
+++ b/app/models/expansion.rb
@@ -1,4 +1,4 @@
 class Expansion < ApplicationRecord
   belongs_to :expansion_type
-  validates :type, presence: true
+  validates :type, :slot, :ports, presence: true
 end
--- a/app/models/expansion.rb
+++ b/app/models/expansion.rb
@@ -1,0 +1,3 @@
+class Expansion < ApplicationRecord
+  belongs_to :expasion_type
+end
--- a/app/models/expansion.rb
+++ b/app/models/expansion.rb
@@ -1,3 +1,4 @@
 class Expansion < ApplicationRecord
   belongs_to :expansion_type
+  validates :type, presence: true
 end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -1,5 +1,8 @@
 
 FactoryBot.define do
+  factory :expansion_type do
+    name "ExpansionDummy"
+  end
   sequence :email do |n|
     "a.scientist.#{n}@liverpool.ac.uk"
   end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -1,8 +1,16 @@
 
 FactoryBot.define do
+  factory :expansion do
+    slot "my-expansion"
+    ports 4
+    expasion_type expansion_type
+    type "expansion"
+  end
+
   factory :expansion_type do
     name "ExpansionDummy"
   end
+
   sequence :email do |n|
     "a.scientist.#{n}@liverpool.ac.uk"
   end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -1,16 +1,5 @@
 
 FactoryBot.define do
-  factory :expansion do
-    slot "my-expansion"
-    ports 4
-    expasion_type expansion_type
-    type "expansion"
-  end
-
-  factory :expansion_type do
-    name "ExpansionDummy"
-  end
-
   sequence :email do |n|
     "a.scientist.#{n}@liverpool.ac.uk"
   end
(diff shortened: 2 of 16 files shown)


bobwhitelock:
Rebased on `develop` so CI should work properly.

spec/factories.rb
```diff
@@ -1,5 +1,16 @@
...
+    slot "my-expansion"
+    ports 4
+    expasion_type expansion_type
+    type "expansion"
+  end
+
+  factory :expansion_type do
+    name "ExpansionDummy"
+  end
+
```
bobwhitelock: I don't think we need these factories at the moment: they're unused and so there's also potential for them to get out-of-sync with the corresponding models, and the flipside of this is we don't want to expend effort maintaining unused code.

Also (not that this matters much since we don't need them), there's a typo in the `expasion_type` field - this should be `expansion_type`, and since this relation refers to the model of the same name you should only need to specify it once.

spec/models/application_record_spec.rb
```diff
@@ -20,14 +20,19 @@ def mock_request_user
...
+
       ActiveRecord::Base.connection.tables.each do |table|
         begin
           klass = table.singularize.camelize.constantize
 
-          # Users are special, they have a relation with a Site but are also
-          # globally available, i.e. able to be read by any other User.
-          next if klass == User
+          # Irregular models:
+          # Users: have a relation with a Site but are also global
```
bobwhitelock: Can we retain the full explanation for why `User`s are an exception (from the left) here please?

spec/models/application_record_spec.rb
```diff
@@ -20,14 +20,19 @@ def mock_request_user
...
+          # Exception: is a base class for STI and is neither
```
bobwhitelock: I think this should say `Expansion` rather than `Exception`. Also the 'and is neither' part doesn't make much sense, can you rephrase what you mean?

spec/models/expansion_type_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe ExpansionType, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
bobwhitelock: Can we follow the suggestion made here and delete these files so they don't spam the logs?

app/models/expansion.rb
```diff
@@ -0,0 +1,3 @@
+class Expansion < ApplicationRecord
+  belongs_to :expansion_type
```
bobwhitelock: Validations for the presence of `slot`, `ports`, and maybe `type` (to avoid creating an `Expansion` which isn't a specific subclass) would be useful here.

spec/models/application_record_spec.rb
bobwhitelock: Don't worry about this comment actually, I'm tweaking how this test generation works and will make this change in the process.

spec/models/application_record_spec.rb
bobwhitelock: As above, don't worry about this comment.

spec/factories.rb
WilliamMcCumstie: The factories have been removed in: https://github.com/alces-software/alces-flight-center/pull/30/commits/08df57611d9eecb6383d1ad621ee7261d8d31698

I have came across that typo else, where it should be fixed now?

app/models/expansion.rb
WilliamMcCumstie: The type validation has been added in: https://github.com/alces-software/alces-flight-center/pull/30/commits/f89e934282edebe62a7b449f7f541d13266ce455.

Will do the others soon

spec/models/application_record_spec.rb
WilliamMcCumstie: Ooops, be you to it: https://github.com/alces-software/alces-flight-center/pull/30/commits/38a9b51007fd3796e947b164f5f9d38822ef102a

app/models/expansion.rb
WilliamMcCumstie: Fixed in: https://github.com/alces-software/alces-flight-center/pull/30/commits/b58f4aca2194bd6fdede4606a8b31d792b81efb9

WilliamMcCumstie:
Ok @bobwhitelock, I have made those changes. Should be ready to merge unless you have got something else?

bobwhitelock:
Nope looks great, thanks!

PR #32, opened 2018-01-08
Improve generated permissions tests

bobwhitelock:
This PR follows on from #30 with a couple of tweaks to the generated model permissions tests. In particular this:

a. always skips the generated test when the model is an STI base class;

b. changes how we iterate through all app models to include STI subclasses, so these are also covered by tests.


WilliamMcCumstie:
This looks much cleaner then my original implementation. Should be good to merge u

bobwhitelock:
Nope happy for you to merge, thanks for checking it :slightly_smiling_face:.

PR #33, opened 2018-01-09
Create ComponentExpansion's based off defaults

WilliamMcCumstie:
When a new component is created, it will automatically create the
default expansions. This is done by reading the DefaultExpansions
from ComponentMake.

The issue with this approach is the :ports, :slot, and :expansion_type
attributes must be manually stated. Because the Expansions use STI,
not all the attributes are ported across.

In particular the id fields must not be copied as this could easily
make an invalid ComponentExpansion. Also the :type field must not be
copied as this will screw around with the STI.

I am open to suggestions on how to do this better. However for the
time being the list needs to be maintained. There is a test for this
behaviour so future changes should be detected.

### DIFF
--- a/app/models/component.rb
+++ b/app/models/component.rb
@@ -9,7 +9,12 @@ class Component < ApplicationRecord
   has_many :asset_record_fields
   has_many :component_expansions
 
+  has_one :component_make, through: :component_group
+  has_many :default_expansions, through: :component_make
+
   validates_associated :component_group, :asset_record_fields
+
+  after_create :create_component_expansions_from_defaults
 
   def asset_record
     # Merge asset record layers to obtain hash for this Component of all
@@ -23,6 +28,16 @@ class Component < ApplicationRecord
   end
 
   private
+
+  def create_component_expansions_from_defaults
+    default_expansions.each do |d|
+      component_expansions.create!(
+        expansion_type: d.expansion_type,
+        slot: d.slot,
+        ports: d.ports
+      )
+    end
+  end
 
   # Method to be called from AdminConfig to format Component asset record for
   # displaying to admins.
--- a/app/models/component.rb
+++ b/app/models/component.rb
@@ -33,11 +33,8 @@ class Component < ApplicationRecord
 
   def create_component_expansions_from_defaults
     default_expansions.each do |d|
-      component_expansions.create!(
-        expansion_type: d.expansion_type,
-        slot: d.slot,
-        ports: d.ports
-      )
+      data = d.slice(:expansion_type, :slot, :ports)
+      component_expansions.create!(**data.symbolize_keys)
     end
   end
 
--- a/app/models/component.rb
+++ b/app/models/component.rb
@@ -12,7 +12,9 @@ class Component < ApplicationRecord
   has_one :component_make, through: :component_group
   has_many :default_expansions, through: :component_make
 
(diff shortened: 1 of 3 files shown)


app/models/component.rb
```diff
@@ -24,6 +29,16 @@ def asset_record
 
   private
 
+  def create_component_expansions_from_defaults
+    default_expansions.each do |d|
+      component_expansions.create!(
+        expansion_type: d.expansion_type,
+        slot: d.slot,
+        ports: d.ports
```
bobwhitelock: You could simplify this, so you only need to specify the keys to copy across once, by using [the `slice` method](http://devdocs.io/rails~5.1/hash#method-i-slice) from active-support.

I would imagine there is some way you could get a list of just the columns which are in both types of `Expansion` and are also not special Rails columns (i.e. which gives `expansion_type`, `slot`, and `ports` but not `id`, `created_at`, `type` etc.), but this would seem very over-complex and convoluted for such a small piece of code, and there is no guarantee that we'd want to continue copying across every new shared field from `DefaultExpansion` to `ComponentExpansion`; so I think just listing out the fields to use is probably the best approach for now.

bobwhitelock:
Rebased on `develop` so this branch only contains the one new commit not already merged into `develop`.

spec/models/component_spec.rb
```diff
@@ -92,4 +92,29 @@
...
+    end
+    let :component_make do
+      create(:component_make, default_expansions: default_expansions)
+    end
+    let :component_group do
+      create(:component_group, component_make: component_make)
+    end
+
+    subject do
+      component_group.components.create!(name: 'test').component_expansions
```
bobwhitelock: This is a good approach to testing this sort of thing: by just creating the object and checking the correct expansions are also created (rather than calling `create_component_expansions_from_defaults` directly) this test should keep working if we change how the expansions are later created :slightly_smiling_face:.

spec/models/component_spec.rb
```diff
@@ -92,4 +92,29 @@
                                          'Comments' => '')
     end
   end
+
+  describe '#create_component_expansions_from_defaults' do
```
bobwhitelock: Might be worth changing this to just be a description of what aspect of `Component`s this is testing (the creation of expansions) rather than to use this method name, this would be slightly more descriptive since this method is private and we're testing the external interface here, and would also mean this wouldn't become out-of-date if this method changed name or changed to an object etc.

spec/models/component_spec.rb
WilliamMcCumstie: :thumbsup:

app/models/component.rb
WilliamMcCumstie: Added the `slice` in: https://github.com/alces-software/alces-flight-center/pull/33/commits/3b60e65dd8895b99957a738c5bb0669cd835c4fc

spec/models/component_spec.rb
```diff
@@ -93,7 +93,7 @@
     end
   end
 
-  describe '#create_component_expansions_from_defaults' do
+  describe "create ComponentExpansion's from ComponentDefault's" do
```
bobwhitelock: FYI you almost always don't need apostrophes to indicate plurals :stuck_out_tongue:

bobwhitelock:
Looks goo, thanks @WilliamMcCumstie!

PR #34, opened 2018-01-09
Dry up the use of `AssetRecordFields`

WilliamMcCumstie:
Based on #33, please merge it first.

This PR consolidates the code for determining the `assert_record` for `ComponentGroup` and `Component`. Instead of maintaining two independent lists of layers that need to be merged, they are now dependent on the higher level.

This is done by creating a new concern and `HasAssetRecordFields` that defines a method `combined_asset_record_fields`. This combined method calls `asset_record_field` on the object and mergers into with the combined assets of it's parent object.

The parent object of `Component` is `ComponentGroup` who in turn has parent `ComponentType`. Please not `ComponentType` does not include the concern as it is the end of the chain.

Note: This was not the original intended implementation. Originally the concern was going to override the  `asset_record_fields` method directly, however this became unworkable at the end. So the original commits look slightly different to the final.

### DIFF
--- a/app/models/component.rb
+++ b/app/models/component.rb
@@ -29,7 +29,7 @@ class Component < ApplicationRecord
     end
   end
 
-  def parent_for_asset_record_fields
+  def asset_record_parent
     component_group
   end
 end
--- a/app/models/component.rb
+++ b/app/models/component.rb
@@ -3,7 +3,7 @@ class Component < ApplicationRecord
   include AdminConfig::Shared::EditableAssetRecordFields
   include ClusterPart
 
-  include HasAssetRecords
+  include HasAssetRecord
 
   belongs_to :component_group
   has_one :component_type, through: :component_group
--- a/app/models/component.rb
+++ b/app/models/component.rb
@@ -29,7 +29,7 @@ class Component < ApplicationRecord
     end
   end
 
-  def parent_asset_record_fields
-    component_group.asset_record_fields
+  def parent_for_asset_record_fields
+    component_group
   end
 end
--- a/app/models/component.rb
+++ b/app/models/component.rb
@@ -3,6 +3,8 @@ class Component < ApplicationRecord
   include AdminConfig::Shared::EditableAssetRecordFields
   include ClusterPart
 
+  include HasAssetRecordFields
+
   belongs_to :component_group
   has_one :component_type, through: :component_group
   has_one :cluster, through: :component_group
@@ -18,17 +20,6 @@ class Component < ApplicationRecord
 
   after_create :create_component_expansions_from_defaults
 
-  def asset_record
-    # Merge asset record layers to obtain hash for this Component of all
-    # asset record fields for this ComponentType; fields set in later layers
-    # will take precedence over those in earlier layers for the same
-    # definition.
-    @asset_record ||=
-      asset_record_layers.reduce({}, :merge).values.map do |field|
-        [field.name, field.value]
-      end.to_h
-  end
-
   private
 
   def create_component_expansions_from_defaults
@@ -38,42 +29,7 @@ class Component < ApplicationRecord
     end
   end
 
-  # Method to be called from AdminConfig to format Component asset record for
-  # displaying to admins.
-  def asset_record_view
-    asset_record.to_json
-  end
-
-  def asset_record_layers
-    # Each entry is a hash of definition ID to asset record field.
-    [
-      empty_asset_record_fields,
-      group_asset_record_fields,
-      component_asset_record_fields,
-    ]
-  end
-
-  def empty_asset_record_fields
-    component_type.asset_record_field_definitions.map do |definition|
-      [
-        definition.id,
-        # Placeholder empty AssetRecordField.
-        AssetRecordField.new(definition: definition, value: ''),
-      ]
-    end.to_h
-  end
-
-  def group_asset_record_fields
-    extract_asset_record_fields(component_group)
-  end
-
-  def component_asset_record_fields
-    extract_asset_record_fields(self)
-  end
-
-  def extract_asset_record_fields(model)
-    model.asset_record_fields.map do |field|
-      [field.definition.id, field]
-    end.to_h
+  def parent_asset_record_fields
+    component_group.asset_record_fields
   end
 end
--- a/app/models/component.rb
+++ b/app/models/component.rb
@@ -3,7 +3,7 @@ class Component < ApplicationRecord
   include AdminConfig::Shared::EditableAssetRecordFields
   include ClusterPart
 
-  include HasAssetRecordFields
+  include HasAssetRecords
 
   belongs_to :component_group
   has_one :component_type, through: :component_group
--- a/app/models/concerns/has_asset_record_fields.rb
+++ b/app/models/concerns/has_asset_record_fields.rb
@@ -3,9 +3,17 @@ module HasAssetRecordFields
   extend ActiveSupport::Concern
 
   def asset_record_fields
-    super
+    hashify_asset_record_fields(parent_asset_record_fields)
+      .merge(hashify_asset_record_fields(super))
+      .values
   end
-  
+
+  def hashify_asset_record_fields(records)
+    (records || []).map do |field|
+      [field.definition.id, field]
+    end.to_h
+  end
+
   def asset_record_layers
     # Each entry is a hash of definition ID to asset record field.
     [
@@ -18,6 +26,7 @@ module HasAssetRecordFields
     extract_asset_record_fields(self)
   end
 
+  # TODO: Remove this
   def extract_asset_record_fields(model)
     model.asset_record_fields.map do |field|
       [field.definition.id, field]
--- a/app/models/concerns/has_asset_record_fields.rb
+++ b/app/models/concerns/has_asset_record_fields.rb
@@ -10,14 +10,14 @@ module HasAssetRecordFields
 
   def asset_record
     @asset_record ||=
-      asset_record_fields.map(&:value).map do |field|
+      combined_asset_record_fields.map do |field|
         [field.name, field.value]
       end.to_h
   end
 
-  def asset_record_fields
+  def combined_asset_record_fields
     hashify_asset_record_fields(parent_asset_record_fields)
-      .merge(hashify_asset_record_fields(super))
+      .merge(hashify_asset_record_fields(asset_record_fields))
       .values
   end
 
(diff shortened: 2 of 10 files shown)


bobwhitelock:
Rebased on latest `develop`.

app/models/concerns/has_asset_record_fields.rb
```diff
@@ -0,0 +1,35 @@
...
+      combined_asset_record_fields.map do |field|
+        [field.name, field.value]
+      end.to_h
+  end
+
+  def combined_asset_record_fields
+    hashify_asset_record_fields(parent_asset_record_fields)
+      .merge(hashify_asset_record_fields(asset_record_fields))
+      .values
+  end
```
bobwhitelock: The approach to forming the asset seems _sort of_ right to me, but I wonder if there's a simpler way. It doesn't quite seem right that we have to 'hashify' the parent asset's fields here, rather than just have the parent do this themselves. 

Possibly by shuffling things around a little bit they would be a bit clearer; in particular with the way things are now it isn't clear to me that things would keep working as expected if another layer was added into forming the asset record (e.g. from `ComponentMake`s).

We also recently discussed how, in order to make editing of the asset record possible, we would need to have assets provide a method to get an array of the `AssetRecordField`s which apply to them (i.e. one `AssetRecordField` for every `AssetRecordFieldDefinition` for the associated `ComponentType`, which may come from the `Component`, the `ComponentGroup`, or be an empty field which doesn't come from the database if neither of these is set).

Combining these two points I could imagine the following methods being defined in this mixin:

- A `combined_asset_record_fields` method (or possibly a better name) which gives a hash of `AssetRecordFieldDefinition` IDs to the corresponding `AssetRecordField` (from this level or above). Once this exists at the top parent level (which it already does, in `ComponentType`, this would just need to be tweaked to include the definition IDs) forming at each further level should be as straightforward as just:

  ```ruby
  parent_asset_record.combined_asset_record_fields.merge(
    asset_record_fields.map {|field| [field.definition.id, field]}.to_h
  )
  ```  

- A changed `asset_record` method which gives this array of `AssetRecordField`s for the current level, which would just be `combined_asset_record_fields.values` once the above exists.

- `asset_record_view` would need to be tweaked to generate the same hash of definition names to field values as it does currently, but using the previous method, I imagine this would be fairly straightforward as just a `map` and a `to_h` across this array.

What do you think of the above suggestions? This seems like it would simplify this mixin and set things up for the asset record editing you're about to add.

app/models/component.rb
```diff
@@ -3,6 +3,8 @@ class Component < ApplicationRecord
   include AdminConfig::Shared::EditableAssetRecordFields
   include ClusterPart
 
+  include HasAssetRecordFields
```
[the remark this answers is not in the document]
WilliamMcCumstie: Renamed: https://github.com/alces-software/alces-flight-center/pull/34/commits/12f0e9a1d09fec03e7b05799d45b263a8dda134f

app/models/component.rb
```diff
@@ -38,42 +29,7 @@ def create_component_expansions_from_defaults
...
-
-  def component_asset_record_fields
-    extract_asset_record_fields(self)
-  end
-
-  def extract_asset_record_fields(model)
-    model.asset_record_fields.map do |field|
-      [field.definition.id, field]
-    end.to_h
+  def parent_for_asset_record_fields
```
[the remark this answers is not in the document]
WilliamMcCumstie: Shortened: https://github.com/alces-software/alces-flight-center/pull/34/commits/496b445b34213901d197915780df299b5bfeec55

app/models/concerns/has_asset_record_fields.rb
WilliamMcCumstie: It will probably be worth while reshuffling the code, however I don't believe this PR is the appropriate  location to do it. Currently there are a few views that are dependent on the `asset_record` being in it's current format.

I am currently consolidating the view code into a partial. This partial will form the basis of the `edit` view as well. Once the code that is dependent on the `asset_record` has been consolidated into the partial, I will have a look at reworking the internals of the concern.

Eventually I intend for the `asset_record` method to return the array and probably remove the `combined_asset_record_fields` method entirely.

WilliamMcCumstie: As far as the hash method is concerned, this is just a convenient way to do the `replace`. However a `reject` would probably work just as easily.

spec/models/concerns/has_asset_record_fields_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe HasAssetRecordFields, type: :model do
+  let :parent do
+    make_obj('parent-asset': 2, 'idx3-parent': 3)
+  end
+
+  subject do
+    make_obj(parent, 'my-asset': 1, 'idx3-child': 3)
```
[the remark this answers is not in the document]
WilliamMcCumstie: The fields have been renamed in: https://github.com/alces-software/alces-flight-center/pull/34/commits/70ccfe062146c7813bb26928ba50391e86dd2eb3

spec/models/concerns/has_asset_record_fields_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def make_obj(parent = nil, **fields)
```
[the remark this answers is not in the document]
WilliamMcCumstie: The purpose of using `double`s is that this module can be used with multiple objects. So I tried to abstract away which object it actually is to the bare minimum required methods.

spec/models/concerns/has_asset_record_fields_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  it 'includes its parent assets' do
+    expect(asset_values).to include('parent-asset')
+  end
+
+  it 'child assets override their parents' do
+    expect(asset_values).to include('idx3-child')
+    expect(asset_values).not_to include('idx3-parent')
+  end
```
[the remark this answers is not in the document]
WilliamMcCumstie: This has already been done, see https://github.com/alces-software/alces-flight-center/blob/05d8afa778ad412bf74b75e65e32a309c073bdd7/spec/models/concerns/has_asset_record_fields_spec.rb#L33

WilliamMcCumstie:
I have made some changes to the `spec` to make it clearer based off the points you made. The reason for the nested `doubles` is because the `AssetRecordFields` are merged depending on the `AssetRecordFieldDefinition.id`.

This should hopefully be clearer in the spec now. Let me know what you think.

app/models/component.rb
```diff
@@ -3,7 +3,7 @@ class Component < ApplicationRecord
   include AdminConfig::Shared::EditableAssetRecordFields
   include ClusterPart
 
-  include HasAssetRecordFields
+  include HasAssetRecords
```
bobwhitelock: Thanks for renaming - sorry for the pedantry but I think this should be `HasAssetRecord` rather than `HasAssetRecords`, as a `Component` etc does not have asset records, but a single asset record as it is a single asset :slightly_smiling_face:.

app/models/concerns/has_asset_record_fields.rb
bobwhitelock: OK fair enough, thanks for explaining. So long as this has been noted for later improvement when doing the editing then I'm happy with this for now.

spec/models/concerns/has_asset_record_fields_spec.rb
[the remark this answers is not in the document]
bobwhitelock: That's a test that the parent's asset record fields are included, however I meant a test that the parent's parent's, i.e. grandparent's, fields are included.

E.g. if there was a relationship between `model1`, `model2`, and `model3`, where `model1.parent_asset_record == model2` and `model2.parent_asset_record == model3`, then I thought a test that `model1.asset_record` includes fields set for `model3` could be useful as it wasn't clear that this would be the case to me.

app/models/component.rb
```diff
@@ -3,6 +3,8 @@ class Component < ApplicationRecord
   include AdminConfig::Shared::EditableAssetRecordFields
   include ClusterPart
 
+  include HasAssetRecords
```
bobwhitelock: Thanks for renaming - sorry for the pedantry but I think this should be `HasAssetRecord` rather than `HasAssetRecords`, as a `Component` etc. does not have asset records, but a single asset record as it is a single asset.

spec/models/concerns/has_asset_records_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def asset_values(obj = subject)
+    obj.combined_asset_record_fields.map(&:value)
+  end
+
+  it 'includes the asset_record_fields for the current layer' do
+    expect(asset_values).to include('subject_asset_field')
+  end
+
+  it 'includes its parent assets' do
```
bobwhitelock: I think this should refer to 'fields' rather than 'assets', as it is the parent asset record fields which are included not the assets themselves.

spec/models/concerns/has_asset_records_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    expect(asset_values).to include('parent_asset_field')
+  end
+
+  it 'subject assets override their parents' do
```
bobwhitelock: As above I think this should refer to 'fields' rather than 'assets'.

app/models/concerns/has_asset_record_fields.rb
WilliamMcCumstie: FYI: This has been addressed in another branch:
https://github.com/alces-software/alces-flight-center/commit/7876e9fd13ba10143fc46e5791af00be44106641

spec/models/concerns/has_asset_record_fields_spec.rb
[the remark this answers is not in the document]
WilliamMcCumstie: Added the test: https://github.com/alces-software/alces-flight-center/pull/34/commits/37892d94f0d4a37407382c0458756da8417cee25

spec/models/concerns/has_asset_records_spec.rb
WilliamMcCumstie: https://github.com/alces-software/alces-flight-center/pull/34/commits/73a5fa939e8c350d2384bac71eb297f676e3e7b1

spec/models/concerns/has_asset_records_spec.rb
WilliamMcCumstie: https://github.com/alces-software/alces-flight-center/pull/34/commits/73a5fa939e8c350d2384bac71eb297f676e3e7b1

app/models/component.rb
WilliamMcCumstie: Renamed in: https://github.com/alces-software/alces-flight-center/pull/34/commits/005750a034a5dd3602f5ec37a59b3087d3f095a5

bobwhitelock:
Wonderful, thanks @WilliamMcCumstie!

PR #36, opened 2018-01-16
Update the asset records from the submitted form

WilliamMcCumstie:
This PR is based on #35, please merge it first.

Once the edit form has been submitted, the `AssetRecordsController` passes it onto the `HasAssetRecord` which does the update. This way the update code can be reused in multiple places.

The logic behind the update is based on the following used by the admin interface. The admin interface has not been switched over to use the module as asset records will likely be edited through the new interface. This means the old admin update code will likely be removed completely at a future date.
https://github.com/alces-software/alces-flight-center/blob/530e4f9f155ce45eaf7c16f64bcad407ba7788e0/app/models/concerns/admin_config/shared/editable_asset_record_fields.rb#L109-L133

Unfortunately the spec for the admin code could not be salvaged as it relies on the admin interface meta-programming. Instead the majority of this PR is focused on recreating the spec on the models themselves. It also has the added benefit of testing `HasAssetRecord` in the manner in which it is used (e.g. `Component`->`ComponentGroup`->`ComponentType`).

### DIFF
--- a/app/controllers/asset_records_controller.rb
+++ b/app/controllers/asset_records_controller.rb
@@ -5,6 +5,7 @@ class AssetRecordsController < ApplicationController
   end
 
   def update
+    asset.update_asset_record(update_param.to_h)
     redirect_to asset
   end
 
--- a/app/controllers/asset_records_controller.rb
+++ b/app/controllers/asset_records_controller.rb
@@ -10,6 +10,11 @@ class AssetRecordsController < ApplicationController
 
   private
 
+  def update_param
+    definition_ids = asset.asset_record.map { |r| r.definition.id.to_s }
+    params.permit(*definition_ids)
+  end
+
   def asset
     @cluster_part || @component_group
   end
--- a/app/models/concerns/has_asset_record.rb
+++ b/app/models/concerns/has_asset_record.rb
@@ -18,6 +18,10 @@ module HasAssetRecord
     parent_asset_record_hash.merge new_asset_record_hash
   end
 
+  def update_asset_record(definition_hash)
+    
+  end
+
   private
 
   def new_asset_record_hash
--- a/app/models/concerns/has_asset_record.rb
+++ b/app/models/concerns/has_asset_record.rb
@@ -32,7 +32,7 @@ module HasAssetRecord
         # When updating a field associated with the asset
         field.value = updated_value
         field.save!
-      elsif updated_value
+      else
         # When updating a higher level field
         create_asset_record_field(field.definition, updated_value)
       end
--- a/app/models/concerns/has_asset_record.rb
+++ b/app/models/concerns/has_asset_record.rb
@@ -22,7 +22,10 @@ module HasAssetRecord
     asset_record.each do |field|
       updated_value = definition_hash[field.definition.id]
       next if field.value == updated_value
-      if field.asset == self
+      if field.asset == self && (updated_value.nil? || updated_value.empty?)
+        # Delete an existing field
+        field.destroy!
+      elsif field.asset
         # When updating a field associated with the asset
         field.value = updated_value
         field.save!
--- a/app/models/concerns/has_asset_record.rb
+++ b/app/models/concerns/has_asset_record.rb
@@ -28,7 +28,7 @@ module HasAssetRecord
       if field.asset == self && (updated_value.nil? || updated_value.empty?)
         # Delete an existing field
         field.destroy!
-      elsif field.asset
+      elsif field.asset == self
         # When updating a field associated with the asset
         field.value = updated_value
         field.save!
--- a/app/models/concerns/has_asset_record.rb
+++ b/app/models/concerns/has_asset_record.rb
@@ -24,7 +24,8 @@ module HasAssetRecord
       next if field.value == updated_value
       if field.asset == self
         # When updating a field associated with the asset
-        raise NotImplementedError
+        field.value = updated_value
+        field.save!
       elsif updated_value
         # When updating a higher level field
         create_asset_record_field(field.definition, updated_value)
--- a/app/models/concerns/has_asset_record.rb
+++ b/app/models/concerns/has_asset_record.rb
@@ -18,9 +18,12 @@ module HasAssetRecord
     parent_asset_record_hash.merge new_asset_record_hash
   end
 
-  def update_asset_record(definition_hash)
+  def update_asset_record(raw_definition_hash)
+    definition_hash = raw_definition_hash.map do |key, value|
+      [key.to_s.to_sym, value]
+    end.to_h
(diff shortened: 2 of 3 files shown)


bobwhitelock:
@WilliamMcCumstie can you rebase this to resolve the conflict with your other PR?

WilliamMcCumstie:
Rebased

app/models/concerns/has_asset_record.rb
```diff
@@ -18,6 +18,27 @@ def asset_record_hash
     parent_asset_record_hash.merge new_asset_record_hash
   end
 
+  def update_asset_record(raw_definition_hash)
+    definition_hash = raw_definition_hash.map do |key, value|
+      [key.to_s.to_sym, value]
+    end.to_h
+    asset_record.each do |field|
+      updated_value = definition_hash[field.definition.id.to_s.to_sym]
+      next if field.value == updated_value
```
bobwhitelock: Not sure if we want this behaviour - even if the current asset record field's value is the same as the updated value, if the field was set at a higher level then won't we still want to create a new field with thuis value for the current level?

If I'm not mistaken in thinking this then we'd want to change this condition to something like `field.asset == self && field.value == updated_value`.

Also either way, since this is also a check for determining how we should handle a change to an asset record field, I think it would read slightly better to include it as another branch in the `if` statement below.

app/models/concerns/has_asset_record.rb
```diff
@@ -18,6 +18,27 @@ def asset_record_hash
...
+      if field.asset == self && (updated_value.nil? || updated_value.empty?)
+        # Delete an existing field
+        field.destroy!
+      elsif field.asset == self
+        # When updating a field associated with the asset
+        field.value = updated_value
+        field.save!
+      else
+        # When updating a higher level field
```
bobwhitelock: Possibly 'When setting a field which is not currently set at this level' would be more descriptive for what we're doing in this case?

app/models/concerns/has_asset_record.rb
```diff
@@ -27,4 +48,11 @@ def new_asset_record_hash
   def parent_asset_record_hash
     asset_record_parent&.asset_record_hash || {}
   end
+
+  def create_asset_record_field(definition, value)
+    asset_record_fields.create!(
+      asset_record_field_definition_id: definition.id,
```
bobwhitelock: When creating a relation between two models you don't need to specify that the `foo_id` field in one is set to `foo.id` - specifying `foo: foo` is sufficient.

Also because of [this line in the field model](https://github.com/alces-software/alces-flight-center/blob/e74950c890e9183996d7115e96ccc986ff7fcd7f/app/models/asset_record_field.rb#L19) it might be sufficient to just use `definition: definition` (not certain of that though).

spec/models/concerns/has_asset_record_spec.rb
```diff
@@ -1,58 +1,194 @@
...
+    it 'subject fields override their parents' do
+      expect(asset_values).to include('subject_override_field')
+      expect(asset_values).not_to include('parent_override_field')
+    end
   end
 
-  it 'subject fields override their parents' do
-    expect(asset_values).to include('subject_override_field')
-    expect(asset_values).not_to include('parent_override_field')
+  describe '#update' do
```
bobwhitelock: The method in the concern is called `update_asset_record` (which is good so we can tell what it's referring to once it's mixed in) - but should change this here I think.

app/controllers/asset_records_controller.rb
```diff
@@ -5,11 +5,17 @@ def edit
   end
 
   def update
+    asset.update_asset_record(update_param.to_h)
     redirect_to asset
   end
 
   private
 
+  def update_param
```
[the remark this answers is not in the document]
WilliamMcCumstie: No longer relevant, the method had to be split between `asset_record_param` and `component_make_id_param`

app/models/concerns/has_asset_record.rb
```diff
@@ -18,6 +18,27 @@ def asset_record_hash
     parent_asset_record_hash.merge new_asset_record_hash
   end
 
+  def update_asset_record(raw_definition_hash)
+    definition_hash = raw_definition_hash.map do |key, value|
```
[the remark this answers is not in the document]
WilliamMcCumstie: https://github.com/alces-software/alces-flight-center/commit/8d426ac3089660f8f0cd3a7b308b119477c94625

app/models/concerns/has_asset_record.rb
WilliamMcCumstie: Obsolete, the logic has changed in a future PR

app/models/concerns/has_asset_record.rb
```diff
@@ -18,6 +18,27 @@ def asset_record_hash
...
+      [key.to_s.to_sym, value]
+    end.to_h
+    asset_record.each do |field|
+      updated_value = definition_hash[field.definition.id.to_s.to_sym]
+      next if field.value == updated_value
+      if field.asset == self && (updated_value.nil? || updated_value.empty?)
```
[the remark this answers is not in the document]
WilliamMcCumstie: Obsolete, we now distinguish between `nil` and empty string

app/models/concerns/has_asset_record.rb
```diff
@@ -18,6 +18,27 @@ def asset_record_hash
...
+        # Delete an existing field
+        field.destroy!
+      elsif field.asset == self
```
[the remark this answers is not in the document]
WilliamMcCumstie: I don't know if it is really worth it. `field` is a local variable within in the loop so it would need to be passed into the method anway

app/models/concerns/has_asset_record.rb
WilliamMcCumstie: https://github.com/alces-software/alces-flight-center/commit/d99b33f1797f9d683378418e17ea5e0e59f91042

app/models/concerns/has_asset_record.rb
WilliamMcCumstie: https://github.com/alces-software/alces-flight-center/commit/c22091a82dff5ea29e9382559ee10ab91c549f68

spec/models/concerns/has_asset_record_spec.rb
WilliamMcCumstie: https://github.com/alces-software/alces-flight-center/commit/4d8d83036dafa2428df16b9b25cd3f4d1318f70d

app/models/concerns/has_asset_record.rb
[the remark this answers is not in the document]
bobwhitelock: Fair enough, makes sense :slightly_smiling_face:

PR #43, opened 2018-01-29
Add page to create/edit/delete ComponentExpansions

WilliamMcCumstie:
Based on #38, please merge it first

This PR is a tad long as it contains a few internal changes to `AssetRecord` code. This is part breaking out the different form elements into `partials` and `decorators` (https://github.com/alces-software/alces-flight-center/commit/04d291c3b35fdfb8e092295b906c14d213d4948f, https://github.com/alces-software/alces-flight-center/commit/5d8a1581527c2cd7adab8c46eb98ab34b3dab31d, https://github.com/alces-software/alces-flight-center/commit/756a27c87ed99295f82dad9ea765cb448c09d7f7) and subsequent changes to how they function. This is to make it easier generate the different common features used by the forms.

There was also a bug fix in which any user could update the asset record due to a bad test. It has been fixed in https://github.com/alces-software/alces-flight-center/commit/0994b82203d49398b3b1a302644d41f2ce56dfb3 and https://github.com/alces-software/alces-flight-center/commit/a067bcd168d7954e9f54db8f7c2abfa7618cc365.

There are various other miscellaneous changes concerning the view and layout, see commit messages for details.

Otherwise the major addition is the expansions table https://github.com/alces-software/alces-flight-center/commit/4f8604d94bbcee0f055b6153d60d9cca6cd34281. It is shown on the component page and performs a bulk edit for all the `ComponentExpansions` associated with the component. It is modelled after the `AssetRecordTable` except it DOES NOT inherit anything from a higher level. `ComponentExpansions` are associated with a single `component` and are independent from the `ComponentGroup/Make`.

The remaining commits add the `update` (https://github.com/alces-software/alces-flight-center/commit/312809d5d2d1d5e1c4da1d80c361899701f08496), `create` (https://github.com/alces-software/alces-flight-center/commit/5ab7fa6e30fb56ee13bee1f48e8156f66984319b), and `destroy` (https://github.com/alces-software/alces-flight-center/commit/641b5b93451fb2696f422077e45473996d6779c2) actions for `ComponentExpansions`.

Note, `edit/update` actions are bulk edits and will update all the expansions associated with the component. The `create/destroy` actions apply to individual components.

PR #46, opened 2018-02-01
Dry up the views for adding the expansions table

WilliamMcCumstie:
This PR is the first step of implementing the expansions. It focuses on updating the views to be able to show the expansions data as it is added/ updated.

It shares a commit history with #43 and should be merged first.

It DRY's up how the header bar of the tables is being created using a partial (`'partials/card_header_nav'`). For the most part, supplying a `title` to the partial is enough for it to be rendered. It does however accommodate list items to be rendered with it, as done in a few locations.

An expansions table has been added to the `component` page. This table has been inspired by `asset_record_table` and shall function in much the same way. In order for the table to be displayed, seed data has been added.

Both `asset_records` and `expansions` table need to toggle the `edit/submit` button in the same fashion. Hence the code to create it has been extracted to a partial.

A bit of work has been done to reduce the logic in the `asset_record_table` partial. This is done by creating a enumerator in a decorator. Do not review this to heavily as it is removed in latter commits.

Please note, the logic for updating the `asset_record` has bee updated as well. Refer to https://github.com/alces-software/alces-flight-center/commit/58b7c7d1314d7aaf0af9e0abb785a589242b08d5

### DIFF
--- a/app/decorators/asset_record_decorator.rb
+++ b/app/decorators/asset_record_decorator.rb
@@ -7,6 +7,38 @@ module AssetRecordDecorator
     h.public_send asset_record_path_method_string, object
   end
 
+  READ_ONLY_MSG = <<-EOF
+    This field can only be set at the individual Component level
+  EOF
+
+  def asset_record_form_inputs
+    Enumerator.new do |memo|
+      object.asset_record.each do |record|
+        is_component_level = (record.definition.level == 'component')
+        options = {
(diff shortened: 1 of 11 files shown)


bobwhitelock:
I should also note that some of these changes obsolete some of my comments on your earlier PRs; some comments may still apply to the refactored version of the code though.

PR #47, opened 2018-02-01
Small bug fix making update asset record admin only

WilliamMcCumstie:
The tests where bad as they where from a non-logged in user which will always fail. However logged in users could access the routes to update the asset record.

This PR fixes the test and updates the routes.

It is based on #46. It should be merged before #43

### DIFF
--- a/spec/features/asset_record.rb
+++ b/spec/features/asset_record.rb
@@ -8,7 +8,8 @@ RSpec.feature 'Asset Record', type: :feature   do
   context 'with an invalid request' do
     it 'errors as a non-admin user' do
       expect do
-        visit edit_component_asset_record_path(component)
+        user = component.site.users.reject(&:admin?).first
+        visit edit_component_asset_record_path(component, as: user)
       end.to raise_error(ActionController::RoutingError)
     end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -29,8 +29,17 @@ Rails.application.routes.draw do
       resources :maintenance_windows, only: :new
     end
 
+    asset_record = Proc.new do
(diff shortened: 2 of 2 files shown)


spec/features/asset_record.rb
```diff
@@ -8,7 +8,8 @@
   context 'with an invalid request' do
     it 'errors as a non-admin user' do
```
bobwhitelock: Updating this text to match the test to mention that specifically this should fail for a Site User would be useful.

spec/features/asset_record.rb
```diff
@@ -8,7 +8,8 @@
...
       expect do
-        visit edit_component_asset_record_path(component)
+        user = component.site.users.reject(&:admin?).first
```
bobwhitelock: I don't think you need the `.reject(&:admin?)` - it is already a guarantee that only non-admin Users can/must have a Site,  so `component.site.users.first` should be sufficient.

WilliamMcCumstie:
Fixed the above issues in:
https://github.com/alces-software/alces-flight-center/commit/211336c7b4db616ce1d910655c2859addf7b592e

spec/features/asset_record.rb
WilliamMcCumstie: Fixed in: https://github.com/alces-software/alces-flight-center/commit/211336c7b4db616ce1d910655c2859addf7b592e

spec/features/asset_record.rb
WilliamMcCumstie: Fixed in: https://github.com/alces-software/alces-flight-center/commit/211336c7b4db616ce1d910655c2859addf7b592e

PR #48, opened 2018-02-02
Add the `edit` and `update` action for `component_expansions`

WilliamMcCumstie:
Based on #47, please merge it first. This PR should be merged and then #43 is the final PR concerning expansions.

A new page has been added to edit component expansions. It preforms a bulk edit in which all the expansions for a single component can be edited at the same time.

All the expansions are otherwise updated independently, meaning if one update fails than the other updates still occur. Any errors will be flashed to the user.

Currently the only constraints are the port has to be a number. The constraints need to be updated to allow nil number of ports.

### DIFF
--- a/app/controllers/component_expansions_controller.rb
+++ b/app/controllers/component_expansions_controller.rb
@@ -13,8 +13,14 @@ class ComponentExpansionsController < ApplicationController
 
   private
 
+  def expansion_errors
+    @errors_in_component_expansion_form_data ||= []
+  end
+
   def update_expansion(expansion)
-    expansion.update!(expansion_param(expansion))
+    unless expansion.update(expansion_param(expansion))
+      expansion_errors.push expansion
+    end
   end
 
   def expansion_param(expansion)
--- a/app/controllers/component_expansions_controller.rb
+++ b/app/controllers/component_expansions_controller.rb
@@ -1,0 +1,4 @@
+class ComponentExpansionsController < ApplicationController
+  def edit
+  end
+end
--- a/app/controllers/component_expansions_controller.rb
+++ b/app/controllers/component_expansions_controller.rb
@@ -1,4 +1,6 @@
 class ComponentExpansionsController < ApplicationController
   def edit
+    @title = "Edit Expansions"
+    @subtitle = @cluster_part.name
   end
 end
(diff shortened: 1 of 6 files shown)


app/views/partials/_expansion_table.html.erb
```diff
@@ -12,14 +13,29 @@
...
+              <td>
+                <%= text_field_tag 'slot' + expansion.id.to_s,
+                                   expansion.slot,
+                                   class: 'form-control' %>
+              </td>
+              <td>
+                <%= text_field_tag 'ports' + expansion.id.to_s,
+                                   expansion.ports,
+                                   class: 'form-control' %>
+              </td>
```
bobwhitelock: I can see why you've done things in this way, creating new param names based on `expansion.id` and then unpacking and using these correctly in the controller, and presumably this works. However this seems quite a complicated/convoluted way to achieve what we want.

Have a look at this video which explains how to do pretty much what we want for multiple updates in a cleaner way: http://railscasts.com/episodes/198-edit-multiple-individually.

It's possibly worth considering switching at some point to do things more like this, both here and for the `AssetRecordField` multiple edit form (where we do something similar AFAICR). Also something worth considering for future kinds of multiple edits, e.g. if we want to support editing multiple expansions across a whole component group.

app/controllers/component_expansions_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def flash_update_error
+    header = "Errors updating expansions:\n"
+    flash[:error] = StringIO.new(header).tap do |io|
+      io.read
+      expansion_errors.each do |expansion|
+        io.puts "#{expansion.expansion_type.name}: #{expansion.errors.full_messages}"
+      end
+      io.rewind
+    end.read
```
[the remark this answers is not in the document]
WilliamMcCumstie: Fixed it by joining an array together. Should be a bit more straight forward. https://github.com/alces-software/alces-flight-center/commit/532204b71f4d0473615c4818978839130cd974c3

app/views/partials/_expansion_table.html.erb
WilliamMcCumstie: The form is working as it is however I agree the other method is cleaner. I have logged it on `trello`

https://trello.com/c/RMRFpoSt/160-remove-manually-portsid-slotid-codec

bobwhitelock: Thanks for logging it

PR #49, opened 2018-02-02
Add data_types to asset records

WilliamMcCumstie:
Based on #43, please merge it first.

This PR implements `data_types` against the asset records. The implementation isn't super straight forward as it affects both `asset_record_fields` and `asset_record_field_definitions`.

The `data_type` itself is stored against the definitions which can be accessed from the fields. As the existing production db does not have values for `data_type`, a default value of `short_text` has been set. When this happens a `deprecation` warning is issued. This means a data migration is not required to update the existing production db.

The first job of the `data_type` is to determine which type of input field (text or area) should be displayed on the edit page. Previously the code to do this was in a helper, but now it has been migrated to two decorators. 

As the field to be displayed is for a specific `asset_record_field` the majority of the input code is contained within the `asset_record_field_decorator`. This sets up the `placeholder`, `value` and `readonly` attributes in the same manner for each record. This decorator then call the `asset_record_field_definition_decorator` which builds the input corresponding to `data_type`.

This two stage process allows code to be shared between all records whilst still generating the different inputs. The default input is a `text_field` as this means the view will always work. It is the job of the validators to check if a `data_type` is valid, not the view.

There are few commits about the implantation of the validation. Initially `STI` was going to be used, however it was scrapped in favour of a concern. There are duplicate commits implementing and re-implementing the validations.

Currently only `short_text` and `long_text` are support types with 50 and 500 character limits respectively. Also within the concern, a list of valid type are generated which is used to validate the definitions. Refer to for more details: https://github.com/alces-software/alces-flight-center/commit/666ccbabc5b5d2e0b0310312a6f5e3f5e02bbb93 and https://github.com/alces-software/alces-flight-center/commit/3ce6ed9bcfe33ac6755aaef9fa7f0ee7dd8f91a8


app/controllers/asset_records_controller.rb
```diff
@@ -5,13 +5,39 @@ def edit
...
+
+  def error_objects
+    @error_objects ||= []
+  end
+
+  def update_asset_record
+    asset.update_asset_record(asset_record_param.to_h)
+         .reject(&:nil?)
+         .reject(&:valid?)
+         .tap { |errors| error_objects.concat errors }
```
bobwhitelock: Rather than mutating the `@error_objects` instance variable and then later accessing this, if you return the errors from this method and then later use the returned errors I think it would be simpler and clearer what's going on. This also makes the dependency between the method using and method setting this variable explicit, and makes it less likely that things will later break due to these getting out-of-sync.

app/decorators/asset_record_field_definition_decorator.rb
```diff
@@ -0,0 +1,23 @@
...
+      text_area(value, options)
+    else
+      text_field(value, options)
+    end
+  end
+
+  private
+
+  def text_area(value, options)
+    options[:style] = "#{options[:style]} height:100px"
```
bobwhitelock: The more typical way to set how big a text area should start off as is to use the `rows` attribute rather than an inline style.

spec/factories.rb
```diff
@@ -70,14 +70,10 @@
     name 'User management'
   end
 
-  factory :asset_record_field_definition do
+  factory :asset_record_field_definition, aliases: [:definition] do
```
bobwhitelock: TIL about `aliases` - nice! :slightly_smiling_face:

spec/models/asset_record_field_definition_spec.rb
```diff
@@ -24,4 +24,63 @@
...
+      end
+
+      it 'is valid' do
+        expect(subject).to be_valid
+      end
+    end
+  end
+
+  describe '#valid?' do
+    context 'with a "forgein_data_type"' do
```
bobwhitelock: `s/forgein/foreign/g`

app/controllers/asset_records_controller.rb
```diff
@@ -5,13 +5,39 @@ def edit
   end
 
   def update
-    asset.update_asset_record(asset_record_param.to_h)
+    update_asset_record
     update_component_make
```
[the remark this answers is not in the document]
WilliamMcCumstie: This issue with using a transaction is the forms are quite long. If the entire transaction is reject than the form inputs are lost. This is unless we prep-populate the form with the (valid?) submitted data.

I have created a card for this: https://trello.com/c/P2zGFeUG/161-wrap-bulk-updates-in-a-transaction

app/controllers/asset_records_controller.rb
```diff
@@ -5,13 +5,39 @@ def edit
...
 
   private
 
+  def flash_errors
+    header = 'The following records failed to update:'
+    error_objects.each_with_object([header]) do |model, array|
+      msg = "#{model.definition.field_name}: #{model.errors.full_messages}"
+      array.push msg
+    end.join("\n")
+       .tap { |full_msg| flash[:error] = full_msg }
```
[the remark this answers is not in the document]
WilliamMcCumstie: Updated.

I'll keep in mind mutating objects in the future

app/controllers/asset_records_controller.rb
WilliamMcCumstie: Fixed

app/decorators/asset_record_field_definition_decorator.rb
WilliamMcCumstie: Fixed

spec/models/asset_record_field_definition_spec.rb
WilliamMcCumstie: Fixed

PR #57, opened 2018-02-06
Migrate the `ComponentGroups` to include `ComponentMake`

WilliamMcCumstie:
When the `db` migration to include `ComponentMake` occurred, it dropped the `component_type` column and added a relationship to `ComponentMake`.

This PR adds some `ComponentMakes` to the `db` and re-establishes the connection to `ComponentGroup`. As the `component_type` column was dropped, these connections needed to be hard coded.

`DefaultExpansions` and `ExpansionTypes` have also been added in this migration.

The development seed data has also been deleted as it is no longer required. Instead the backups of the production database can be used for development.


bobwhitelock:
LGTM, thanks, just tweaked a couple of names and will merge in a minute.

bobwhitelock:
Not sure why CI is failing, will investigate tomorrow.

PR #58, opened 2018-02-06
Final tweaks for ComponentMake and Expansions

WilliamMcCumstie:
Reasonable small PR based on #57 

It contains a few final commits of requested changes. Please refer to the commit messages for more details. Once this PR has been merged, then we can look at releasing it.

### DIFF
--- a/app/controllers/component_expansions_controller.rb
+++ b/app/controllers/component_expansions_controller.rb
@@ -7,7 +7,7 @@ class ComponentExpansionsController < ApplicationController
       flash[:success] = msg
     else
       expansion_errors.push expansion
-      flash_errors 'Could not add the epansion'
+      flash_errors 'Could not add the expansion'
     end
     redirect_back fallback_location: @cluster_part
   end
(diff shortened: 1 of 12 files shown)


spec/models/expansion_spec.rb
```diff
@@ -10,10 +10,19 @@
...
-    it 'must have ports and slot set' do
-      expect(subject).to include(:ports)
+    it 'must have slot set' do
       expect(subject).to include(:slot)
     end
   end
+
+  describe 'ports' do
+    # NOTE: DefaultExpansion is used as a stand in as the type must be set
+    # However it is testing the underlining Expansion validation
```
bobwhitelock: @WilliamMcCumstie for future reference you could do something like this

```ruby
[:default_expansion, :component_expansion].each do |expansion_type|
  it 'defaults to zero if not provided' do
    expansion = create(expansion_type, ports: '')
    expect(expansion).to be_valid
    expect(expansion.ports).to eq 0
  end
```

if you want to ensure this covers both possibilities.

app/controllers/component_expansions_controller.rb
```diff
@@ -7,7 +7,7 @@ def create
       flash[:success] = msg
     else
       expansion_errors.push expansion
-      flash_error 'Could not add the epansion'
+      flash_errors 'Could not add the epansion'
```
bobwhitelock: `s/epansion/expansion/`

bobwhitelock:
Rebased on `develop`.

PR #59, opened 2018-02-07
Make Sentry DSN configuration optional.

jamesremuscat:
If a DSN isn't set (e.g. for running in development) then Sentry won't report errors. This is probably what we want when running on dev machines.


bobwhitelock:
Wonderful, thanks @jamesremuscat - just rebased this on the latest `develop` and will merge in a minute.

PR #60, opened 2018-02-08
Update flight branding

bobwhitelock:
Many thanks to @jamesremuscat for doing this, and his perseverance when making everything look correct. I've rebased this branch on `develop` ready to be merged and to remove my first junk commit.


bobwhitelock:
We have a slight issue with the navbars with this branch - the top (black) bar now appears to work correctly, but the lower (blue) bar is not showing any of the links which should appear. I will investigate - this may delay rolling this out @mjtko @jamesremuscat.

PR #62, opened 2018-02-08
Create the `ClusterLog` model

WilliamMcCumstie:
This will allow "engineers" to log events against a cluster. It is directly linked to the cluster and thus the logs do not need to be associated with an open case. The `engineer` can be any flight centre admin user.

The `ClusterLog` can also be associated with cases. This is a many-many relationship to allow for flexibility. The only constraint is all the `cases` belong to the same cluster as the log itself. This is enforced at the model level.

A `cluster_log.related_ticket_numbers` method has not been included at this time. It is essence an alias of `cluster_log.cases.map(&:rt_ticket_id)` which is of similar length. The `RT` ids are likely going to be a view issue and not something required at the model level. Any form that needs to display the `RT_id` will also need the `id` which is the actual primary key.

### DIFF
--- a/app/models/cluster_log.rb
+++ b/app/models/cluster_log.rb
@@ -1,5 +1,6 @@
 
 class ClusterLog < ApplicationRecord
+  has_one :site, through: :cluster
   belongs_to :cluster
   belongs_to :engineer, class_name: 'User', foreign_key: "user_id"
   has_and_belongs_to_many :cases
--- a/app/models/cluster_log.rb
+++ b/app/models/cluster_log.rb
@@ -4,5 +4,14 @@ class ClusterLog < ApplicationRecord
   belongs_to :engineer, class_name: 'User', foreign_key: "user_id"
 
   validates :details, presence: true
+  validate :engineer_is_a_admin
+
+  private
+
+  def engineer_is_a_admin
+    unless engineer&.admin?
+      errors.add(:engineer, 'must be an admin')
+    end
+  end
 end
 
--- a/app/models/cluster_log.rb
+++ b/app/models/cluster_log.rb
@@ -20,7 +20,7 @@ class ClusterLog < ApplicationRecord
   def cases_belong_to_cluster
     cases.each do |kase|
       unless kase.cluster == cluster
-        msg = "##{kase.rt_ticket_id} is for a different cluster"
+        msg = "ticket ##{kase.rt_ticket_id} is for a different cluster"
         errors.add(:cases, msg)
       end
     end
--- a/app/models/cluster_log.rb
+++ b/app/models/cluster_log.rb
@@ -1,2 +1,8 @@
+
 class ClusterLog < ApplicationRecord
+  belongs_to :cluster
+  belongs_to :engineer, class_name: 'User', foreign_key: "user_id"
+
+  validates :details, presence: true
 end
+
--- a/app/models/cluster_log.rb
+++ b/app/models/cluster_log.rb
@@ -6,12 +6,22 @@ class ClusterLog < ApplicationRecord
 
   validates :details, presence: true
   validate :engineer_is_a_admin
+  validate :cases_belong_to_cluster
 
   private
 
   def engineer_is_a_admin
     unless engineer&.admin?
       errors.add(:engineer, 'must be an admin')
+    end
+  end
+
+  def cases_belong_to_cluster
+    cases.each do |kase|
+      unless kase.cluster == cluster
+        msg = "##{kase.rt_ticket_id} is for a different cluster"
+        errors.add(:cases, msg)
+      end
     end
   end
 end
--- a/app/models/cluster_log.rb
+++ b/app/models/cluster_log.rb
@@ -1,0 +1,2 @@
+class ClusterLog < ApplicationRecord
+end
--- a/app/models/cluster_log.rb
+++ b/app/models/cluster_log.rb
@@ -2,6 +2,7 @@
 class ClusterLog < ApplicationRecord
   belongs_to :cluster
   belongs_to :engineer, class_name: 'User', foreign_key: "user_id"
+  has_and_belongs_to_many :cases
 
   validates :details, presence: true
   validate :engineer_is_a_admin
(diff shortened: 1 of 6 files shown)


db/migrate/20180208121324_add_cluster_log.rb
```diff
@@ -0,0 +1,11 @@
+class AddClusterLog < ActiveRecord::Migration[5.1]
+  def change
+    create_table :cluster_logs do |t|
+      t.text :details, null: false
+      t.references :cluster, null: false
+      t.references :user, null: false
+    end
+
+    create_join_table :cases, :cluster_logs, column_options: { null: true }
```
bobwhitelock: Do we want the `column_options: { null: true }` here? I may be mistaken, but doesn't this mean we could have entries in this table where either `cluster_log_id` or `case_id` is `null`? But I can't imagine a situation where we would want this to happen.

spec/models/cluster_log_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    subject { create(:cluster_log) }
+
+    def create_case(case_subject)
+      kase = create(:case, cluster: subject.cluster, subject: case_subject)
+      subject.cases << kase
+      subject.reload
+    end
+
+    it 'can have multiple cases' do
+      case_msgs = ['first', 'second', 'third']
```
bobwhitelock: Calling these `case_subjects` might be more logical, since that's what they're used as - what's a `msg`?

spec/models/cluster_log_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      case_msgs.each { |m| create_case(m) }
+      expect(subject).to be_valid
+      expect(subject.cases.map(&:subject)).to contain_exactly(*case_msgs)
+    end
+
+    it 'validates that all cases belong to its cluster' do
+      ['other-case1', 'other-case2', 'other-case3'].each do |case_msg|
```
bobwhitelock: As above, calling these `case_subject` seems a more logical name.

bobwhitelock: Also: why do we need to create 3 Cases here? It would seem we would have as robust a test with one Case which belongs to the correct Cluster and one which doesn't, but it would be more obvious what's going on.

spec/models/cluster_log_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        create_case(case_msg)
+      end
+      other_cluster = create(:cluster, name: 'other-cluster')
+      bad_case = create(:case, cluster: other_cluster, subject: 'Bad Case')
+      subject.cases << bad_case
+      subject.reload
+      expect_single_error subject, "#{bad_case.rt_ticket_id}"
```
bobwhitelock: Here and in the previous test, can we split the lines up into 'setup', 'execute', and 'assert' sections to better indicate what we're doing?

app/models/cluster_log.rb
```diff
@@ -0,0 +1,29 @@
...
+  def engineer_is_a_admin
+    unless engineer&.admin?
+      errors.add(:engineer, 'must be an admin')
+    end
+  end
+
+  def cases_belong_to_cluster
+    cases.each do |kase|
+      unless kase.cluster == cluster
+        msg = "##{kase.rt_ticket_id} is for a different cluster"
```
bobwhitelock: Won't the double hash here mean this will read like `#10000 is for a different cluster`? Also could be worth prefixing this error message to indicate what the `kase.rt_ticket_id` means, as this may be non-obvious by the time it is shown to a user, e.g.

```ruby
msg = "ticket #{kase.rt_ticket_id} is for a different cluster"
```

WilliamMcCumstie: The `##{...}` is intentional to indicate it is a number. The next line does automatically preface the error with `cases`. However I will look at making it more descriptive

spec/models/cluster_log_spec.rb
WilliamMcCumstie: I have been calling it `msg` for two reasons. `case_subject` is to close to `subject` which was causing confusion when I initially had it that way.

Also `expect(subject.cases.map(&:subject)).to contain_exactly(*case_subjects)` was to long to fit on one line

spec/models/cluster_log_spec.rb
WilliamMcCumstie: Rearranged slightly, however this spec is basically all `setup` except the single assertion at the end.

spec/models/cluster_log_spec.rb
WilliamMcCumstie: Removed

db/schema.rb
```diff
@@ -72,13 +72,28 @@
...
+  create_table "cluster_logs", force: :cascade do |t|
+    t.text "details", null: false
+    t.bigint "cluster_id", null: false
+    t.bigint "user_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["cluster_id"], name: "index_cluster_logs_on_cluster_id"
+    t.index ["user_id"], name: "index_cluster_logs_on_user_id"
+  end
+
```
bobwhitelock: @WilliamMcCumstie This PR now includes the schema changes but the migration itself appears to have got lost - do you know where it is?

WilliamMcCumstie: Hopefully in my reflog :|

spec/models/cluster_log_spec.rb
bobwhitelock: Fair enough, just thought it was a minor point of possible future confusion but not a big issue.

bobwhitelock:
Thanks for making the changes, this looks good to merge to me as soon as we've tracked down that missing migration.

One minor point for future reference: in your new commits you link to https://github.com/alces-software/alces-flight-center/pull/62/files#r167026963, presumably to indicate the original reason for the change (which would be useful) - however this is the URL for all files changed in this PR. It would be slightly more useful to link to the specific comment you're addressing, which you can get the URL for by copying the link to the timestamp for that comment (i.e. the `5 days ago` in the comment header).

WilliamMcCumstie:
Ok found the missing `migration` must  of got lost during a rebase. Ohh turns out I linked to the commit and not the comment. Opps

Should be ready to merge

bobwhitelock:
Great, thanks!

PR #64, opened 2018-02-09
Create the route, view and controller for ClusterLog

WilliamMcCumstie:
This PR is based on #62, please merge it first

Most of this PR is reasonable straight forward. There was no major refactoring required in implementing the view. The `index` page is used to display all the logs for a cluster. `admins` also have an additional form (on the same page) to add a new log.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,7 +13,8 @@ class ApplicationController < ActionController::Base
 
   def error_flash_models(models, header)
     flash[:error] = header + "\n" + models.map do |model|
-      block_given? ? (yield model) : model.errors.full_messages
+      prefix = block_given? ? ((yield model).to_s + ': ') : ''
+      prefix + model.errors.full_messages.to_s
     end.join("\n")
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,7 +12,9 @@ class ApplicationController < ActionController::Base
   rescue_from ReadPermissionsError, with: :not_found
 
   def error_flash_models(models, header = 'Errors:')
-    flash[:error] = header + "\n" + (models.map { |m| yield m }).join("\n")
+    flash[:error] = header + "\n" + models.map do |model|
+      block_given? ? (yield model) : model.errors.full_messages
+    end.join("\n")
   end
 
   private
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,7 +11,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from ReadPermissionsError, with: :not_found
 
-  def error_flash_models(models, header = 'Errors:')
+  def error_flash_models(models, header)
     flash[:error] = header + "\n" + models.map do |model|
       block_given? ? (yield model) : model.errors.full_messages
     end.join("\n")
--- a/app/controllers/cluster_logs_controller.rb
+++ b/app/controllers/cluster_logs_controller.rb
@@ -1,0 +1,5 @@
+class ClusterLogsController < ApplicationController
+  def index
+    @title = 'Logs'
+  end
+end
--- a/app/controllers/cluster_logs_controller.rb
+++ b/app/controllers/cluster_logs_controller.rb
@@ -7,7 +7,12 @@ class ClusterLogsController < ApplicationController
   end
 
   def create
-    p log_params
+    new_log = @cluster.cluster_logs.create(**log_params)
+    if new_log.valid?
+      flash[:success] = 'Added new log entry'
+    else
+      error_flash_models [new_log], 'Could not add log entry'
+    end
     redirect_back fallback_location: @cluster
   end
 
@@ -16,7 +21,7 @@ class ClusterLogsController < ApplicationController
   def log_params
     params.require(:cluster_log)
           .permit(:details, case_ids: [])
-          .to_h
+          .to_h.symbolize_keys
           .merge(engineer: current_user)
           .tap do |h|
             h[:case_ids] = h[:case_ids].reject(&:blank?)
--- a/app/controllers/cluster_logs_controller.rb
+++ b/app/controllers/cluster_logs_controller.rb
@@ -1,8 +1,8 @@
 class ClusterLogsController < ApplicationController
   def index
     @title = 'Logs'
+    @new_log = ClusterLog.new
     @logs = @cluster.cluster_logs
-    @new_log = @cluster.cluster_logs.new
     @cases = @cluster.cases.order(created_at: :desc)
   end
 
--- a/app/controllers/cluster_logs_controller.rb
+++ b/app/controllers/cluster_logs_controller.rb
@@ -7,5 +7,19 @@ class ClusterLogsController < ApplicationController
   end
 
   def create
+    p log_params
+    redirect_back fallback_location: @cluster
+  end
+
+  private
+
+  def log_params
+    params.require(:cluster_log)
+          .permit(:details, case_ids: [])
+          .to_h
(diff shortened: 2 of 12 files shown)


bobwhitelock:
@WilliamMcCumstie this now has conflicts after merging the previous PR - would you be able to rebase?

WilliamMcCumstie:
@bobwhitelock, rebased

app/controllers/application_controller.rb
```diff
@@ -12,7 +12,9 @@ class ApplicationController < ActionController::Base
   rescue_from ReadPermissionsError, with: :not_found
 
   def error_flash_models(models, header = 'Errors:')
-    flash[:error] = header + "\n" + (models.map { |m| yield m }).join("\n")
+    flash[:error] = header + "\n" + models.map do |model|
+      block_given? ? (yield model) : model.errors.full_messages
```
bobwhitelock: This line seems dense, and if I delete the `block_given? ? (yield model) : ` part then no tests fail - is it actually needed anywhere?

app/controllers/cluster_logs_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  private
+
+  def log_params
+    params.require(:cluster_log)
+          .permit(:details, case_ids: [])
+          .to_h.symbolize_keys
+          .merge(engineer: current_user)
+          .tap do |h|
+            h[:case_ids] = h[:case_ids].reject(&:blank?)
+          end
```
bobwhitelock: If I delete this `tap` block no tests fail - is it needed? If so, what's the purpose and can we add test coverage?

app/controllers/cluster_logs_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  def index
+    @title = 'Logs'
+    @new_log = ClusterLog.new
+    @logs = @cluster.cluster_logs
+    @cases = @cluster.cases.order(created_at: :desc)
+  end
+
+  def create
+    new_log = @cluster.cluster_logs.create(**log_params)
+    if new_log.valid?
```
bobwhitelock: Validations are run for all normal ActiveRecord methods, and the `!`-less versions return whether they succeeded, so I believe you can replace this with `if new_log` and achieve the same effect.

app/controllers/cluster_logs_controller.rb
```diff
@@ -0,0 +1,30 @@
+class ClusterLogsController < ApplicationController
+  def index
+    @title = 'Logs'
+    @new_log = ClusterLog.new
+    @logs = @cluster.cluster_logs
+    @cases = @cluster.cases.order(created_at: :desc)
+  end
+
+  def create
+    new_log = @cluster.cluster_logs.create(**log_params)
```
bobwhitelock: Don't think you need the `**` here.

app/decorators/case_decorator.rb
```diff
@@ -10,6 +10,7 @@ def maintenance_window_form_info
       "Created by #{user.name}"
     ].join(' | ')
   end
+  alias_method :rt_long_text, :maintenance_window_form_info
```
bobwhitelock: Do we need 2 names for the same method? We could name this something more general to what it's doing and then use it in multiple places with the same name - what about `case_select_details`?

app/views/cluster_logs/index.html.erb
```diff
@@ -0,0 +1,52 @@
...
+          <%= f.text_area :details, class: 'form-control'%>
+        </div>
+        <div class='form-group'>
+          <%= f.label :case_ids, 'Related Ticket(s)' %>
+          <%= f.collection_select :case_ids,
+                                  @cases.decorate,
+                                  :id,
+                                  :rt_long_text,
+                                  {},
+                                  { multiple: true } %>
```
bobwhitelock: This select is quite ugly atm - can we add some Bootstrap classes to it to style it like the rest of our forms?

config/routes.rb
```diff
@@ -79,6 +80,7 @@
     resources :clusters, only: :show do
       resources :cases, only: :new
       resources :consultancy, only: :new
+      resources :cluster_logs, path: 'logs', only: :index
```
bobwhitelock: We could call this resource `logs`, and then we wouldn't need to specify a new path for it and the route helper would be the slightly nicer `cluster_logs_path`. This would also be slightly more generic for when we add 'component logs' in the near future.

app/controllers/application_controller.rb
WilliamMcCumstie: So the yield is needed by the views so they can construct an error string and return it. By removing everything would still pass but we can't customise the message

app/controllers/cluster_logs_controller.rb
WilliamMcCumstie: Removed

app/controllers/cluster_logs_controller.rb
WilliamMcCumstie: Removed

app/decorators/case_decorator.rb
WilliamMcCumstie: Renamed

app/views/cluster_logs/index.html.erb
WilliamMcCumstie: Fixed

config/routes.rb
WilliamMcCumstie: I will old off doing this for now. When it comes to implementing the "ComponentLog", I'll do a migration to rename the table `logs`

app/controllers/cluster_logs_controller.rb
WilliamMcCumstie: I don't believe that will work. `create!` and `update!` will error if the validation fails. Also `update` returns whether the validation succeeded. `create` return the actual object (`ClusterLog` in this case), otherwise you wouldn't have the object that was just created.

WilliamMcCumstie:
@bobwhitelock, addressed your issues and added a spec for adding a new log

app/controllers/application_controller.rb
bobwhitelock: OK - I think this function and its usage still seems over-complicated however. To simplify it I think you could:

- remove the default for `header` as it's never used anywhere;

- since in the only places you use this function with a block: https://github.com/alces-software/alces-flight-center/blob/3665ed548ddd882ee4de3e6b79ecf0b84b59fd63/app/controllers/asset_records_controller.rb#L17 and https://github.com/alces-software/alces-flight-center/blob/3665ed548ddd882ee4de3e6b79ecf0b84b59fd63/app/controllers/component_expansions_controller.rb#L52 you always call `model.errors.full_messages` anyway, I think you could simplify those places by changing this line something like this:

  ```ruby
  prefix = block_given? ? yield model + ': ' : ''
  prefix + model.errors.full_messages
  ```

Then the places where you call this could be changed to just generate the parts unique to each of them (the prefix).

config/routes.rb
bobwhitelock: :+1:

bobwhitelock:
Thanks for the changes, I've added a couple of responses for you to look at above.

app/controllers/cluster_logs_controller.rb
WilliamMcCumstie: Ahh good point, that's a nice way around it. However instead of using `log.new` I think `log.build` is more appropriate. The distinguish probably does matter to much in this case, but `build` puts the new log into the collection.

app/controllers/application_controller.rb
WilliamMcCumstie: Updated

app/controllers/cluster_logs_controller.rb
bobwhitelock: Good idea :+1:

PR #74, opened 2018-02-20
Replace `ClusterLog` with `Log` and implement "component logs"

### LINKED ISSUE #23: Add Component logs
We'd like to add a service record to each component page, editable only by alces admins, but viewable off each component by the user.

A service record page would show each service record entry and allow new service records to be added. 

A service records is

Date/Time entered | Engineer (alces admin who added the record) | Details (enough for a brief description but keep the page fairly neat) | Related Ticket Numbers

WilliamMcCumstie:
Fixes #23 

This PR implements logs that can be made against a component. To prevent a second table from being made, the `ClusterLog` model has been renamed to `Log` which required renaming the variables, adding a migration and updating the routes. A data migration is not required as the db migration maintains the integrity of the indices.

A lot of work has been done to clean up the feature spec for `Log`. This spec insures the add log form and controller operates correctly for the different log types (component and cluster).

The changes to the `Log` model are relatively minor. A new optional `component` field has been added. The component is then validated to belong to the cluster. If a `Log` is initialized without a cluster but with a component, than the cluster is automatically inferred from the component.

When creating a new component log, the form does not submit the component id. This is because the component it is being added to can be inferred from the `URI` path. This has been done by setting the `scope` instance variable. However the code to do this is temporary as the scope shall be set in the `Application` controller (eventually).

NOTE: the component instance variable is also set in the `ApplicationController`. See https://github.com/alces-software/alces-flight-center/pull/74/commits/1d88c3148d009cda79f03674dd1dad71c25c71e0 and https://github.com/alces-software/alces-flight-center/pull/74/commits/8800f039ee7bdf86500014ad03f5ebda00cd333e for details. This maybe replaced by the scope eventually.


spec/factories.rb
```diff
@@ -1,6 +1,6 @@
 
 FactoryBot.define do
-  factory :cluster_log do
+  factory :log do
```
bobwhitelock: Forgot to mention this on the previous log-related PR, but can we move this down the bottom of this file so it reads from the most core models (sites, users etc) downwards? (I know `factory_bot` generates them up the top).

spec/features/log_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      log_details = fill_details_input
+      log = submit_log
+
+      expect(log).not_to be_nil
+      expect(log.details).to eq(log_details)
+      expect(log.engineer).to eq(engineer)
+      expect(log.cases).to be_blank
+    end
+
+    it 'can add multiple tickets/cases to the view' do
```
bobwhitelock: Not sure what 'to the view' means here - might be more accurate to say `it 'can associate multiple tickets/cases with the log' do`?

spec/models/log_spec.rb
```diff
@@ -62,5 +62,28 @@ def create_case(case_subject)
       expect_single_error subject, "#{bad_case.rt_ticket_id}"
     end
   end
+
+  it 'does not require a component' do
+    expect(create(:log).component).to be_nil
```
bobwhitelock: This test could maybe be a bit clearer. In general a standard `FactoryBot` factory (one with the same name as the model) should set the minimum amount of properties for the model to be valid, but these should not be explicitly depended upon in tests as this is then coupling the tests to a definition in a distant factory definitions file, which could make things hard to debug if the factory or model later change.

In this case you're depending on the `log` factory not setting a component, and testing that this is the case here - you're not testing whether or not the created model is valid, apart from implicitly as the `create` will fail if it is not.

A clearer way to test what I think you want to test here might be the following:

```ruby
it 'does not require a component' do
  log = build(:log, component: nil)
  expect(log).to be_valid
end
```

spec/factories.rb
WilliamMcCumstie: Moved

spec/features/log_spec.rb
WilliamMcCumstie: Updated

spec/models/log_spec.rb
WilliamMcCumstie: Yeah, that's much clearer. Thanks for the tip. I've made the change.

WilliamMcCumstie:
Rebased on develop

bobwhitelock:
Can I :shipit: :question:

WilliamMcCumstie:
:shipit:!

Issue #42, opened 2018-01-29
Maintenance period extensions


ste78:
We need to extend the maintenance functionality with date ranges, so that we can book something out for maintenance (and apply for authorisation from the customer) between a start time and end time, rather than immediately.

bobwhitelock:
@ste78 I have some questions on how you think this should work. Currently the way maintenance windows work is:

- an admin requests maintenance for a particular component/service/cluster and in association with a particular case;

- a user confirms this;

- the admin later ends the maintenance.

At each step of this process a comment is left on the corresponding ticket indicating what happened and any action needed by the user. The component/service/cluster is also considered to be 'under maintenance' for the period between confirmation by the user and maintenance being ended by the admin.

In the new system:

1. Should the requested start and end times always be required to be set by the admin when requesting maintenance? Or can these optionally be unset, in which case the previous behaviour will presumably remain?

2. If these can be unset, should this be possible independently? So maintenance could have a requested start time but will end whenever this is triggered by the admin (or vice versa)?

3. If maintenance windows do have start and/or end times set, should this automatically indicate the item is under maintenance/maintenance has ended, and make a comment on the original ticket indicating this?

4. For the initial version of this feature, do we need to support users altering the requested maintenance start and end times?

ste78:
1. All maintenance requests should have a start / end time. The current ones without times are not really that useful. 
2. can’t be unset
3. yes if been approved, if it’s not approved before the start date then the maintenance requests becomes expired a comment sent out at the start time saying that ‘the maintenance request was not authorised by the user in time’
4. Nope, editing not required. but we can cancel one and make another with different dates though?

bobwhitelock:
Thanks for the previous answers @ste78 - one new thing which has now come up: with the change to maintenance being for scheduled times, is there still a need for the quick maintenance request buttons for cases (the spanner shown alongside cases which immediately requests maintenance for the associated cluster/component/service)?

Now that all maintenance needs a requested start and end these will no longer be able to immediately request the maintenance, and would need to just become links to start requesting maintenance with the given cluster/component/service and case already selected, so I'm not sure how useful they would be over just creating maintenance in the more general way.

ste78:
yep agreed, for other coming functionality l’m also thinking maybe we drop that feature. can we discuss on monday IRL?

bobwhitelock:
[...]

PR #75, opened 2018-02-21
New maintenance window states and UI adaptations

bobwhitelock:
I'm aware this PR is very large; it's grown to include a lot of things and has gone through a few iterations in the process. I was going to keep working on this until it reached a mostly complete state, but since you said you'd like to see PRs as they go in I'm submitting this now before it gets any larger. I can try to split it up more if you do want to have a closer look, but alternatively it might be simpler for us both if you want to have a skim and then I'll make future PRs smaller and more reviewable.

In brief, this PR consists of:

- Adding/adapting all MaintenanceWindow states (we are now using the [state_machines](https://github.com/state-machines/state_machines) Gem and some related Gems) which will likely be needed in the new maintenance process, and the transitions between them.

- Possibly most useful for understanding the above is this diagram, generated using the [state_machines-graphviz](https://github.com/state-machines/state_machines-graphviz) Gem, which shows the possible states a MaintenanceWindow can now be in and how the transitions can occur.

- Each time any of these transitions occur an appropriate comment will be left on the ticket for the Case related to that maintenance; see https://github.com/alces-software/alces-flight-center/blob/7d184753981ddc4a6d9cef9d149d8f48390c57aa/app/models/maintenance_notifier.rb for details of how these comments currently look (though I plan to update these and make them more consistent before releasing this).

- The UI has also been updated to account for the above changes, as well as to allow setting a `requested_start` and `requested_end` when creating a MaintenanceWindow. However this only means the parts of the UI which remain needed still work, it does not mean the whole maintenance process works as I still need to handle automatic transitions in a later PR (`request`, `confirm`, `reject` and `cancel` are user-initiated state transitions, whereas `start`, `end`, and `expire` should happen automatically after a certain time is reached).

This is part of implementing https://trello.com/c/L2GTfslB/173-maintenance-period-extensions / #42.


bobwhitelock:
Thanks for reviewing.

The `MaintenanceWindowStateTransition` model is automatically created and used by the [state_machines-audit_trail](https://github.com/state-machines/state_machines-audit_trail) Gem. It represents and logs every time a MaintenanceWindow transitions from one state to another. The reason for using this is that we want to track and display in various places when these events happen, as well as who initiated the transition if this is done by a user (e.g. so we can show things like 'Engineer William McCumstie requested maintenance 1 day ago' and 'Steve User confirmed maintenance 1 hour ago').

Initially I was tracking the transitions with many fields in the main MaintenanceWindow model like `requested_at` and `requested_by`; however this led to an explosion of complexity with appropriately updating and validating all these fields, and doing this all consistently (you can see this in the earlier commits in this PR). Instead, I realized these fields are essentially all a similar kind of thing, and so tracking them in independent models which a MaintenanceWindow has many of makes things much clearer and easier to work with. Also see https://github.com/alces-software/alces-flight-center/pull/75/commits/1b1ad26f4a20d25eae56f715476a3721f2c6b488 and https://github.com/alces-software/alces-flight-center/pull/75/commits/6a1b6617892494122fb9877a35771af55bdc7f55 for more details on this.

PR #77, opened 2018-02-21
Only include unfinished maintenance windows when getting open maintenance windows

bobwhitelock:
This is a small, self-contained PR which builds upon #75 (so merge that first). It just updates the `open_maintenance_windows` method, provided for all models which can have maintenance windows, to account for and exclude MaintenanceWindows in any of the 'finished' states which are now possible; some minor other tweaks need to do this are also included.

### DIFF
--- a/app/models/cluster.rb
+++ b/app/models/cluster.rb
@@ -13,7 +13,6 @@ class Cluster < ApplicationRecord
   has_many :cases
   has_many :credit_deposits
   has_many :credit_charges, through: :cases
-  has_many :maintenance_windows
   has_many :logs, dependent: :destroy
 
   validates_associated :site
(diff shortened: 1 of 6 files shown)


WilliamMcCumstie:
Rebased on develop

WilliamMcCumstie:
LGTM

bobwhitelock:
I think this line got kept in the rebase: https://github.com/alces-software/alces-flight-center/blob/e180353a079aa3748950e5874b24f6023a80d962/app/models/cluster.rb#L16 (it's now been moved to the concern so no longer needs to be in the core model). I'll rebase the original version again to remove this; then will merge when CI passes for that.

bobwhitelock:
OK merged, thanks for reviewing :shipit:

PR #78, opened 2018-02-22
Maintenance progression service

bobwhitelock:
Based on #77. This PR implements a new object, `ProgressMaintenanceWindow`, which takes a MaintenanceWindow as an argument and then progresses its state or not as needed when its `progress` method is called, based on the current time and the `requested_start` and `requested_end` of the maintenance. This method also returns an appropriate message about this change or lack of it, so this can later be logged. Actually using this new object anywhere is still to come in a future PR.

### DIFF
--- a/app/services/progress_maintenance_window.rb
+++ b/app/services/progress_maintenance_window.rb
@@ -1,0 +1,34 @@
+
+ProgressMaintenanceWindow = Struct.new(:window) do
+  def progress
+    if end_time_passed? && started?
+      window.end!
+    elsif start_time_passed?
+      start_or_expire_if_unstarted
+    end
+  end
+
+  private
+
+  delegate :confirmed?, :started?, to: :window
+
+  def end_time_passed?
+    window.requested_end.past?
+  end
+
+  def start_time_passed?
+    window.requested_start.past?
+  end
+
+  def start_or_expire_if_unstarted
+    if confirmed?
+      window.start!
+    elsif unconfirmed?
+      window.expire!
+    end
+  end
+
+  def unconfirmed?
+    window.new? || window.requested?
+  end
+end
--- a/app/services/progress_maintenance_window.rb
+++ b/app/services/progress_maintenance_window.rb
@@ -1,16 +1,20 @@
 
 ProgressMaintenanceWindow = Struct.new(:window) do
   def progress
-    if end_time_passed? && started?
-      window.end!
-    elsif start_time_passed?
-      start_or_expire_if_unstarted
-    end
+    progress_if_needed || unprogressed_message
   end
 
   private
 
   delegate :confirmed?, :started?, to: :window
+
+  def progress_if_needed
+    if end_time_passed? && started?
+      progress_state(:ended)
+    elsif start_time_passed?
+      start_or_expire_if_unstarted
+    end
+  end
 
   def end_time_passed?
     window.requested_end.past?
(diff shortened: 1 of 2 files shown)


WilliamMcCumstie:
Re-based on develop

app/services/progress_maintenance_window.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  private
+
+  delegate :confirmed?, :started?, to: :window
+
+  def progress_if_needed
+    if end_time_passed? && started?
+      progress_state(:ended)
+    elsif start_time_passed?
+      start_or_expire_if_unstarted
```
WilliamMcCumstie: It is a bit confusing having an `or` and `if` embedded in this method within another `if-elsif` statement. The logic control would be clearer if all the logic was within a single `if-elsif-else` statement.

app/services/progress_maintenance_window.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def start_time_passed?
+    window.requested_start.past?
+  end
+
+  def start_or_expire_if_unstarted
+    if confirmed?
+      progress_state(:started)
+    elsif unconfirmed?
```
WilliamMcCumstie: This `if-elsif` statement looks a bit odd. Surely the maintenance is either `confirmed` or `unconfirmed`? I assume there is a reason this has been implemented this way but it isn't obvious to me.

From the looks of it, I would have thought this could be done with an `if-else` statement.

spec/services/progress_maintenance_window_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+        states.each do |state|
+          window = create_window(state: state)
+
+          result = described_class.new(window).progress
+
+          expect(window.state.to_sym).to eq state
+          test_progression_message(
+            message: result,
+            window: window,
+            expected: "remains #{state}"
```
WilliamMcCumstie: It maybe worth extracting this code possible to a function. From what I can see, the only difference is the `expected: <some message>` line and this spec runs in a loop.

Otherwise this spec is almost exactly the same as the `progress` spec

spec/services/progress_maintenance_window_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    context 'when just requested_start passed' do
+      def create_window(state:)
+        create(
+          :maintenance_window,
+          state: state,
+          requested_start: 1.hours.ago,
+          requested_end: DateTime.current.advance(days: 1),
+          component: component,
+        )
+      end
```
WilliamMcCumstie: This method definition is essentially a repeat to the one above. Thus any future changes will need to be duplicated. It also implies these are different when in reality they are not.

Instead the method should be defined higher up as:
```
def create_window(:state)
  create( ... inputs ..., request_start: request_start, request_end: request_end)
end
```

Then withing the contexts define `let` blocks for the differences
```
let :request_start { 1.hours.ago }
let :request_end { DateTime.current.advance(days: 1) }
```

This way the spec highlights that the `create_window` method is the same BUT the `request_start` and `request_end` are different.

spec/services/progress_maintenance_window_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    context 'when requested_start and requested_end in future' do
+      def create_window(state:)
+        create(
+          :maintenance_window,
+          state: state,
+          requested_start: DateTime.current.advance(days: 1),
+          requested_end: DateTime.current.advance(days: 2),
+          component: component,
+        )
```
WilliamMcCumstie: See comment below for `create_window`

spec/services/progress_maintenance_window_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+
+    context 'when requested_start and requested_end passed' do
+      def create_window(state:)
+        create(
+          :maintenance_window,
+          state: state,
+          requested_start: 2.hours.ago,
+          requested_end: 1.hours.ago,
+          component: component,
```
WilliamMcCumstie: See comment above for `create_window`

bobwhitelock:
@WilliamMcCumstie Thanks for the review and the useful comments. I've made a new PR addressing your inline comments at #84, to make things simpler since these files have changed further since this PR was made. I'll merge this one through now but please make any additional comments you have on the comments on that PR.

To address your general comments:

> My major comment about the ProgressMaintenanceWindow struct is the logic is quite complicated. There are nested logic structures that seem to be missing else options where I would have expected to see one.

Hopefully this has been improved with the changes I've made to address your inline comments as well as related refactoring, which should make clearer that essentially `progress` switches on some different conditions where we should progress the window, or falls back to `unprogressed_message` if none of these are met.

> Also is there any particular reason this is a `struct` instead of a `class`?

Only because a `Struct` is a slightly simpler way to express a class which just takes in some arguments and stores them as instance variables with accessors, and this indicates that we don't need any special handling of the arguments in `initialize`.

> The `spec` looks good except there seems to be repeated code. It maybe worth extracting some of it into shared methods and using `let` in the context to configure the differences.

Should be improved by addressing the inline comments.

PR #82, opened 2018-02-23
Fix issue with implicit state transitions in ProgressMaintenanceWindow

bobwhitelock:
This PR/commit fixes an issue I noticed in ProgressMaintenanceWindow while doing other things, see https://github.com/alces-software/alces-flight-center/commit/84d39848bce777609885d8141a6f27054dd80632 for details. Based on #81.


bobwhitelock:
@WilliamMcCumstie by merging this without rebasing it on the latest `develop` you've duplicated some commits from the PR it is based on in the history, which makes it slightly more difficult to follow. Not a major deal but can you try to rebase things on the latest `develop` before merging. You shouldn't need to wait for CI to run again when you do this as the code will be literally identical, just with less duplicated commits in the history. Thanks :slightly_smiling_face:

PR #83, opened 2018-02-23
Migrate existing maintenance windows to new format

bobwhitelock:
This PR adds a data migration to update all existing MaintenanceWindows to be in the new format Flight Center expects. The main commit doing this is https://github.com/alces-software/alces-flight-center/commit/43962b0adc9156d3b93f000016f049c6d7630ae1, and in order to do this I also needed to add a way to skip commenting on an RT ticket when a MaintenanceWindow transitions, which is done in the earlier commits. Based on #82.

### DIFF
--- a/app/models/maintenance_window.rb
+++ b/app/models/maintenance_window.rb
@@ -11,6 +11,8 @@ class MaintenanceWindow < ApplicationRecord
   validates_presence_of :requested_start
   validates_presence_of :requested_end
 
+  attr_accessor :skip_comments
+
   state_machine initial: :new do
     audit_trail context: :user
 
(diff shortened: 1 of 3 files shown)


bobwhitelock:
@WilliamMcCumstie forget to mention, but this should address the issue you encountered the other day (https://alces.slack.com/archives/C72GT476Y/p1519392024000334) by making the existing maintenance windows valid for the new format.

WilliamMcCumstie:
Rebased on develop

WilliamMcCumstie:
LGTM. I can't give anything more detailed as I'm not super familiar with what makes a valid maintenance window. However this looks like it can be merged

PR #84, opened 2018-02-26
Address PR 78 comments

bobwhitelock:
This PR addresses comments on #78. Based on #83.

### DIFF
--- a/app/services/progress_maintenance_window.rb
+++ b/app/services/progress_maintenance_window.rb
@@ -9,13 +9,20 @@ ProgressMaintenanceWindow = Struct.new(:window) do
   delegate :confirmed?, :started?, to: :window
 
   def progress_if_needed
-    if end_time_passed? && started?
-      transition_state(:end)
-    elsif start_time_passed? && confirmed?
-      transition_state(:start)
-    elsif start_time_passed? && unconfirmed?
-      transition_state(:expire)
+    if required_transition_event
+      transition_state(required_transition_event)
     end
+  end
+
+  def required_transition_event
+    @required_transition_event ||=
+      if end_time_passed? && started?
+        :end
+      elsif start_time_passed? && confirmed?
+        :start
+      elsif start_time_passed? && unconfirmed?
+        :expire
+      end
   end
 
   def end_time_passed?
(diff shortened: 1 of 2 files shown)


WilliamMcCumstie:
Rebased on develop

app/services/progress_maintenance_window.rb
```diff
@@ -9,15 +9,22 @@ def progress
...
-      transition_state(:start)
-    elsif start_time_passed? && unconfirmed?
-      transition_state(:expire)
+    if required_transition_event
+      transition_state(required_transition_event)
     end
   end
 
+  def required_transition_event
+    @required_transition_event ||=
```
WilliamMcCumstie: TIL that you don't need to go `||= begin .... end` for multi-lined assignment

app/services/progress_maintenance_window.rb
```diff
@@ -33,7 +33,7 @@ def start_time_passed?
     window.requested_start.past?
   end
 
-  def unconfirmed?
+  def not_yet_confirmed?
```
WilliamMcCumstie: Ohh I see, it specifically refers to the `unconfirmed` state. In that case `unconfirmed_state?` would also do. Up to you if you want to make that change

WilliamMcCumstie: Alternatively a comment on what `unconfirmed?` means should suffice. It is a bit awkward having `confirmed` and `unconfirmed` as possible states, however it is the nature of the beast.

app/services/progress_maintenance_window.rb
bobwhitelock: You can only do this in this case as `if ... [many elsifs] ... else ...` is a single expression which returns the value from whichever branch got evaluated; you'd still need the `begin ... end` if you needed intermediate assignments, however in that case it would often be nicer to split out a method rather than use `begin` anyway.

app/services/progress_maintenance_window.rb
bobwhitelock: I prefer my name as it specifically refers to the states when the MaintenanceWindow has not been confirmed but still could be, whereas there are states where the maintenance window could be considered 'unconfirmed' and could no longer be confirmed (e.g. 'expired', when it has not been confirmed by the requested start time, or 'rejected', when a site contact rejects the maintenance). Glad that makes things clearer though.

bobwhitelock:
Merged; thanks for all the reviewing :shipit:

PR #85, opened 2018-02-26
Implement a scope to be used by the nav bar

WilliamMcCumstie:
Previously a bunch of navigation variables need to be set in order for the navigation bar to constructed. The problem with this is it makes it hard for other controllers to know if the nav variable is the model for the current page.

Instead a `scope` variable has been set instead. This makes it clear which model belongs to the current page and can be used by other controller (e.g. `LogController` - eventually). This PR implements the `scope` and updates the navigation variable spec.

Almost all the extraneous variables have been removed with exception of the `cluster_part` which is still required.

This PR does completely break the nav bar. This will be fixed in the next PR which focuses on refactoring it.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -44,20 +44,19 @@ class ApplicationController < ActionController::Base
     case request.path
     when /^\/sites/
       id = params[:site_id] || params[:id]
-      @site = Site.find(id) if current_user.admin?
+      @scope = @site = Site.find(id) if current_user.admin?
     when /^\/clusters/
       id = params[:cluster_id] || params[:id]
-      @cluster = Cluster.find(id)
+      @scope = @cluster = Cluster.find(id)
     when /^\/components/
       id = params[:component_id] || params[:id]
-      @component = Component.find(id)
-      @cluster_part = @component
+      @scope = @cluster_part = @component = Component.find(id)
     when /^\/component-groups/
       id = params[:component_group_id] || params[:id]
-      @component_group = ComponentGroup.find(id)
+      @scope = @component_group = ComponentGroup.find(id)
     when /^\/services/
       id = params[:service_id] || params[:id]
-      @cluster_part = Service.find(id)
+      @scope = @cluster_part = Service.find(id)
     end
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -41,8 +41,6 @@ class ApplicationController < ActionController::Base
   def define_navigation_variables
     return unless current_user
 
-    @site = current_user.site
-
     case request.path
     when /^\/sites/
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -30,47 +30,40 @@ def assign_current_user
   end
 
   def current_site
-    @site
+    @site || @scope.site
```
bobwhitelock: Site does actually implement a `site` method so we can check the site for all models related to a site in a generic way, so this could be simplified to just `@scope.site`.

app/controllers/application_controller.rb
```diff
@@ -30,47 +30,40 @@ def assign_current_user
...
   end
 
   # From https://stackoverflow.com/a/4983354/2620402.
   def not_found
     raise ActionController::RoutingError.new('Not Found')
   end
 
+  def scope_id_params(id_method)
```
bobwhitelock: This only returns the value for a single param, so would probably be better named as `scope_id_param`.

app/controllers/application_controller.rb
```diff
@@ -30,47 +30,40 @@ def assign_current_user
...
+               id = scope_id_params(:component_id)
+               @component = @cluster_part = Component.find(id)
+             when /^\/component-groups/
+               id = scope_id_params(:component_group_id)
+               @component_group = ComponentGroup.find(id)
+             when /^\/services/
+               id = scope_id_params(:service_id)
+               @service = @cluster_part = Service.find(id)
+             else
+               @site = current_user.site unless current_user.admin?
```
bobwhitelock: Rather than having `@site` (and `@scope` when it's a site) be set in two distinct places, it might be clearer if this was always done in this `else` branch. This might look something like this:

```ruby
else
  @site = if request.path =~ /^\/sites/
            id = scope_id_params(:site_id)
            Site.find(id)
          elsif current_user.contact?
            current_user.site
          end
end
```

You could also split a method out for the `if` block if you wanted.

app/controllers/application_controller.rb
WilliamMcCumstie: Removed

app/controllers/application_controller.rb
WilliamMcCumstie: Renamed

app/controllers/application_controller.rb
WilliamMcCumstie: Consolidated down into the `else`. I haven't broken it out into a method as it is only a few lines of code. At this point it is still reasonable easy to follow

WilliamMcCumstie:
Fixed those issues

bobwhitelock:
Thanks for fixing, this can be merged once the next PR is ready to be merged (not merging now to avoid `develop` being broken for any length of time).

PR #86, opened 2018-02-26
Clean up the rendering of the nav bar

WilliamMcCumstie:
Based on #85, either merge it first or skip it and merge this one. This PR fixes the errors in the previous PR.

The rendering of the nav bar has been completely overhauled to rely on the scope. This means other controllers can now set instance variables without having to worry about affecting the navigation. The nav bar has been broken out into its own partial template. The nav links also have there own partial as this prevents duplication of the code.

When rendering a nav link, the `text`, `path` and `icon` for the link are required. Alternatively, a `model` can be provided which will implicitly set the `text` and `path`.

Things get a bit tricky when trying to set the active link. Previously this was done with a bunch of logic, which isn't idea as it is more likely to be bug prone. Instead the last link is always rendered as the active link (which should always be the case).

This is a bit difficult as the number of links are not know in advance. So instead, the first pass creates an array of procs. These procs contain the code required to render each link. Then it becomes a trivial case of rendering the last proc as the active link.

### DIFF
--- a/app/views/partials/_nav_bar.html.erb
+++ b/app/views/partials/_nav_bar.html.erb
@@ -15,10 +15,9 @@
     <ul class="navbar-nav mr-auto">
       <% if signed_in? %>
         <% if current_user.admin? %>
-          <%= render 'partials/nav_link',
-                     text: 'All Sites',
-                     path: root_path,
-                     icon: 'fa-globe' %>
+          <% add_nav_link_proc text: 'All Sites',
+                               path: root_path,
+                               icon: 'fa-globe' %>
         <% end %>
 
         <% if @scope %>
@@ -28,34 +27,33 @@
                              else
                                root_path
                              end %>
-          <%= render 'partials/nav_link',
-                     text: site_obj.name,
-                     path: path_for_site,
-                     icon: 'fa-institution' %>
+          <% add_nav_link_proc text: site_obj.name,
+                               path: path_for_site,
+                               icon: 'fa-institution' %>
 
           <% cluster_obj = model_from_scope :cluster %>
           <% if cluster_obj %>
-            <%= render 'partials/nav_link',
-                      text: cluster_obj.name,
-                      path: cluster_obj,
-                      icon: 'fa-server' %>
+            <% add_nav_link_proc text: cluster_obj.name,
+                                 path: cluster_obj,
+                                 icon: 'fa-server' %>
           <% end %>
 
           <% component_group_obj = model_from_scope :component_group %>
           <% if component_group_obj %>
-            <%= render 'partials/nav_link',
-                       text: component_group_obj.name,
-                       path: component_group_obj,
-                       icon: 'fa-cubes' %>
+            <% add_nav_link_proc text: component_group_obj.name,
+                                 path: component_group_obj,
+                                 icon: 'fa-cubes' %>
           <% end %>
 
           <% if @cluster_part %>
-            <%= render 'partials/nav_link',
-                       text: @cluster_part.name,
-                       path: @cluster_part,
-                       icon: 'fa-cube' %>
+            <% add_nav_link_proc text: @cluster_part.name,
+                                 path: @cluster_part,
+                                 icon: 'fa-cube' %>
           <% end %>
         <% end %>
+      <% end %>
+      <% nav_link_procs.each do |nav_proc| %>
+        <%= nav_proc.call false %>
       <% end %>
     </ul>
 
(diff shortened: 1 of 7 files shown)


app/views/partials/_nav_bar.html.erb
```diff
@@ -0,0 +1,77 @@
...
+    aria-controls="navbarSupportedContent"
+    aria-expanded="false"
+    aria-label="Toggle navigation"
+  >
+    <span class="navbar-toggler-icon"></span>
+  </button>
+
+  <div class="collapse navbar-collapse" id="navbarSupportedContent">
+    <ul class="navbar-nav mr-auto">
+      <% if signed_in? %>
```
bobwhitelock: Not checked, but you may be able to remove this check as I think `current_user` and `@scope` can only be set if a user is signed in, so this is not guarding against anything which would happen otherwise (though you'll probably need to change the next line to `current_user&.admin?`.

app/views/partials/_nav_bar.html.erb
```diff
@@ -0,0 +1,77 @@
...
+            <% add_nav_link_proc model: component_group_obj,
+                                 icon: 'fa-cubes' %>
+          <% end %>
+
+          <% if @cluster_part %>
+            <% add_nav_link_proc model: @cluster_part,
+                                 icon: 'fa-cube' %>
+          <% end %>
+        <% end %>
+      <% end %>
```
bobwhitelock: This seems like too much and too complex logic to embed in a view. As a first step, you could remove wrapping every line in `<%` and `%>`, since no templating is being done here other than evaluating a lot of Ruby, and instead have a single pair of these wrapping this entire section.

As a more involved refactoring, it may be better to pull out some/all of this logic to a helper or decorator. However this step probably isn't necessary for now.

app/views/partials/_nav_bar.html.erb
```diff
@@ -0,0 +1,77 @@
...
+      <% nav_link_procs[0..-2].each do |nav_proc| %>
+        <%= nav_proc.call %>
+      <% end %>
+      <%= nav_link_procs.last&.call(active: true) %>
```
bobwhitelock: Rather than slicing the original array and then special-casing the last item here, would something like this work:
```ruby
<% nav_link_procs.each do |nav_proc| %>
  <% active = nav_proc == nav_link_procs.last %>
  <%= nav_proc.call(active: active) %>
<% end %>
```

app/views/partials/_nav_link.html.erb
```diff
@@ -0,0 +1,8 @@
+<% path ||= model %>
+<% text ||= model.name %>
+<li class='nav-item'>
+  <%= link_to (raw("<span class='fa #{icon}'></span> ") + text),
+              path,
+              class: "nav-link #{ 'nav-link--active' if (active ||= false) }"
```
bobwhitelock: With the change to iterating through the `nav_link_procs` above I believe you will be able to remove the `if (active ||= false)` for `active` and just use `if active`, as it will always be set.

app/views/partials/_nav_bar.html.erb
WilliamMcCumstie: I agree, I have put them into a single set of erb tags. The pile of logic is a artefact of how these links where originally being rendered. However that has become unworkable due to how the active link needs to be set. I have created a card for it.

https://trello.com/c/HCPHLefw/185-refactor-the-nav-links-to-use-a-helper

app/views/partials/_nav_bar.html.erb
WilliamMcCumstie: Removed

app/views/partials/_nav_bar.html.erb
WilliamMcCumstie: Fixed

app/views/partials/_nav_link.html.erb
WilliamMcCumstie: So I still have `false` as the default as there is another use of the partial lower in the template. Also I intend to use the `nav-link` partial for the cluster tabs

WilliamMcCumstie:
@bobwhitelock, addressed your comments

app/views/partials/_nav_link.html.erb
WilliamMcCumstie: Actually this has been implemented in the end. It will be in a latter PR however.

bobwhitelock:
LGTM, thanks :shipit:

PR #87, opened 2018-02-27
Move all shared_examples to `spec/support/`

bobwhitelock:
Very minor change; just submitting on its own to keep this distinct from the next PR to come.

PR #88, opened 2018-02-27
Scheduled maintenance window progression

bobwhitelock:
This PR:

- implements a Rake task which will progress all unfinished maintenance windows;

- sets up a generic Rake task which will run every minute, to run the above and any other future similar tasks we might have;

- sets up a means of testing these and other Rake tasks;

- documents how to install the `cron` Rake task for our next deploy.

Based on #87.


WilliamMcCumstie:
Rebased on develop

PR #89, opened 2018-02-28
Implement `respond_to?` for MaintenanceWindows

bobwhitelock:
This PR implements `respond_to?` for the `MaintenanceWindow` class, which is the best practice when `method_missing` is implemented and helps avoid potential odd problems when these give inconsistent results (which is what prompted me to do this after having issues in a later PR). Some related refactoring is also included. Based on #88.

### DIFF
--- a/app/models/maintenance_window.rb
+++ b/app/models/maintenance_window.rb
@@ -78,19 +78,12 @@ class MaintenanceWindow < ApplicationRecord
   end
 
   def method_missing(symbol, *args)
-    super unless symbol =~ /^([a-z]+)_(at|by)$/
-    state = $1.to_sym
-    property = $2.to_sym
-    super unless self.class.possible_states.include?(state)
-    last_transition_property(state: state, property: property)
+    query = TransitionQuery.parse(symbol)
+    query ? query.value_for(self) : super
   end
 
   def respond_to?(symbol, include_all=false)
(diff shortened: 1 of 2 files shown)


WilliamMcCumstie:
Rebased on develop

app/models/maintenance_window.rb
```diff
@@ -78,11 +78,12 @@ def associated_cluster
...
-    super unless symbol =~ /^([a-z]+)_(at|by)$/
-    state = $1.to_sym
-    property = $2.to_sym
-    super unless self.class.possible_states.include?(state)
-    last_transition_property(state: state, property: property)
+    query = TransitionQuery.parse(symbol)
+    query ? query.value_for(self) : super
+  end
+
+  def respond_to?(symbol, include_all=false)
```
WilliamMcCumstie: Typically you shouldn't override `respond_to?`. It can have some weird side affects. Instead `respond_to_missing?` should be defined with `method_missing`. This way you don't have to worry about the `super`.

https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding

bobwhitelock:
Thanks for the review and pointing out that issue - I've made that change but it will be coming in a future PR to avoid conflicts since I've made other changes since making this one, and I'll merge this one through now.

PR #90, opened 2018-02-28
Update maintenance table

bobwhitelock:
This PR updates the maintenance table shown on the Cluster page to show new data and support new actions, following implementing the new maintenance window system. This includes:

- showing the requested time period for each requested/scheduled maintenance window;

- showing a cancel maintenance button to admins;

- showing a reject maintenance button to contacts;

- various related refactoring behind the scenes.

Based on #89.

Issue #73, opened 2018-02-20
Compulsory maintenance


ste78:
Allow an admin to override the user confirmation stage when making a maintenance request in exceptional cases

bobwhitelock:
It is currently possible (and will remain possible in the new maintenance system being worked on at the moment unless we add explicit handling to prevent this) for an admin to confirm maintenance themselves. This means an admin can already bypass the user confirmation step of a maintenance request if needed.

Is this sufficient for the initial version of this feature? If not, how should this differ from what is already possible?

ste78:
Yeah thats what's currently happening, but I'm not keen on how it looks, it looks like we are just skipping getting confirmation from the end user - It needs to look more intentional... "Steve Norledge has marked this node down for compulsory maintenance" ie visibly skip the confirmation procedure.

bobwhitelock:
Yep OK, makes sense - do you want the confirmation skipping to happen when initially submitting the maintenance request form or as an extra step once this has been done and the request made on the original ticket (or both)?

bobwhitelock:
Have as checkbox when request maintenance

bobwhitelock:
Start time is then irrelevant; maintenance started as soon as form submitted - could have this selected instead of start date.

bobwhitelock:
This was done in https://github.com/alces-software/alces-flight-center/pull/139.

PR #91, opened 2018-03-01
Validate MaintenanceWindow transition user

bobwhitelock:
This PR validates that the user who initiated a MaintenanceWindow transition is of the appropriate type, i.e. only admins can request or cancel maintenance and only users can reject it (confirming maintenance will be only allowed by users later, once https://github.com/alces-software/alces-flight-center/issues/73 has been implemented). Also included is some refactoring of the transition tests and a fix so MaintenanceWindows do not blow up when an invalid transition occurs.

Based on #90.


WilliamMcCumstie:
Rebased on develop

PR #95, opened 2018-03-02
Redirect users seamlessly if not already signed in

WilliamMcCumstie:
Addresses: https://trello.com/c/l5msCYDI/116-if-try-to-visit-page-while-logged-out-redirect-to-page-after-login

If a user tries to visit a page but isn't signed in, they are automatically sent to the sign in page before being redirected back to the initial page they tried to visit.

This is done by embedding the page they initially tried to visit into the sign in form (defaulting to the root site). This is then used by the redirect at the end of the sign in process. 

To achieve this, the clearance `sessions` controller needed to be overridden. For more information on this, please see: https://github.com/thoughtbot/clearance#controllers


spec/features/asset_record_spec.rb
```diff
@@ -6,11 +6,13 @@
   let :component_group { create(:component_group) }
 
   context 'with an invalid request' do
+    # NOTE: This test is a bit funky with the new redirect if signed out
+    # As the spec is technically signed out, it is redirecting to sign in
+    # instead of erroring
```
bobwhitelock: I'm don't think I follow this - I can see that this test passes with your changes, and that no `RoutingError` is raised from the test, however if I manually try to visit an asset record edit page as a non-admin user I still get the `RoutingError`, which is what previously happened and what I would expect/want to happen. Do you know why there is the difference in behaviour between the test and manually interacting with things? Ideally we would have the test match the behaviour we want to happen, rather than test something different which seems to somewhat correspond with what we want.

spec/requests/request_spec.rb
```diff
@@ -20,10 +20,12 @@ def show
...
     end
 
+    # No user now means they are logged out. So instead the assertion
+    # should be if the redirect occurred correctly
     it 'returns nil when in request and no current user' do
       get request_test_path
-
-      expect(assigns(:current_user)).to be nil
+ 
+      assert_redirected_to root_path(path: request_test_path)
```
bobwhitelock: Don't think I follow why we need this change - these tests are for the `Request` module, which provides access to the `current_user` throughout the app when we are within a request. This particular test was to check that `Request.current_user` is `nil` when we are within a request but no user is signed in. However it now doesn't check this and checks an unrelated thing, whether we are redirected, which passes but does not test what we originally wanted.

I think this test should still test the same thing as before, and likely should look the same as it did before, but we may need some changes in other places to allow it to still work. Possibly the definition of `request_test` in the routes should be adjusted so that it is matched before every remaining route is matched for the redirect?

WilliamMcCumstie:
I was contemplating instead of redirecting users to `/?path=original_visited_path` we could render the `sign_in` page directly when visiting `/original_visited_path`. This removes the round trip to the server whilst also preserving the path in the advent of a re-render.

This would require accepting a `POST` to any path of `non_signed_in` users but it would also preserve where the users are making the request from.

bobwhitelock:
Thanks for you work on this before @WilliamMcCumstie :slightly_smiling_face:. This has been superseded now we've switched to SSO, and the original card has now been addressed in https://github.com/alces-software/alces-flight-center/pull/355.

PR #96, opened 2018-03-05
Fix invalid cases when model support type changed

bobwhitelock:
This PR fixes an issue where Cases could become invalid if their associated model's `support_type` is later changed; this could then cause an error in the UI in some circumstances. See https://github.com/alces-software/alces-flight-center/commit/ec819c7f74a4610e849e71c4278c40d4b7a060b3 for an explanation of the issue and fix; the other commits here are just refactoring following this change. The diff here is quite large but most of this is just churn due to renaming/splitting some files and changing indentation.

Trello: https://trello.com/c/EFipx2DP/186-fix-error-when-admin-views-all-cases-page-and-some-cases-have-since-become-invalid-due-to-their-associated-models-support-type-c.

### DIFF
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -36,6 +36,14 @@ class Case < ApplicationRecord
     presence: true,
     inclusion: {in: TICKET_STATUSES}
 
+  # Only validate Issue relationship on create, as the Issue must be allowed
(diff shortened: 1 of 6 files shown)


WilliamMcCumstie:
LGTM

PR #97, opened 2018-03-05
Remove seed data

bobwhitelock:
I thought we'd removed this a while ago, but that must have only been
some of the seed data. This is now unneeded and likely out-of-date; we
now set the database up by importing recent production data so we can
develop with realistic data.


WilliamMcCumstie:
I think the development seed data may have been removed.

PR #99, opened 2018-03-12
Automatically update all VCR fixtures

bobwhitelock:
Nothing interesting here, just updating the VCR fixtures, which automatically expire once a month. You'll need to have this merged, and other things rebased on it, for all tests to run successfully (unless you have valid credentials for everything set up, in which case it shouldn't matter).


WilliamMcCumstie:
Ahh `vcr` my old friend

PR #100, opened 2018-03-12
Track `requested_start` and `requested_end` in MaintenanceWindow transitions

bobwhitelock:
The values for `requested_start` and `requested_end` may soon potentially change for a MaintenanceWindow at different points; in particular a site contact will soon be able to  change these values when they confirm some maintenance. Therefore we now want to track these values as they change in the MaintenanceWindow transition model so that we do not lose this potentially useful information.

One small step towards https://trello.com/c/k81APmbi/191-expired-maintenance-should-still-show-in-table-users-can-always-change-requested-start-and-end-date-when-confirming; based on #99.

### DIFF
--- a/app/models/maintenance_window.rb
+++ b/app/models/maintenance_window.rb
@@ -16,7 +16,7 @@ class MaintenanceWindow < ApplicationRecord
   attr_accessor :skip_comments
 
   state_machine initial: :new do
-    audit_trail context: :user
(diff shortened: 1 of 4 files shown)


WilliamMcCumstie:
Rebased on develop

WilliamMcCumstie:
LGTM

PR #101, opened 2018-03-12
MaintenanceWindow `requested_start` and `requested_end` validations

bobwhitelock:
This PR adds various validations for the MaintenanceWindow `requested_start` and `requested_end` validations, to ensure certain invariants about these properties hold:

- `requested_start` should always be before `requested_end`;

- `requested_start` and `requested_end` cannot be in the past, apart from in certain 'finished' states for a MaintenanceWindow to avoid every saved MaintenanceWindow eventually becoming invalid which would likely cause us future problems.

Adding these validations now became particularly important as non-admin users will soon be able to change these fields, so we want to be more robust when given bad data to make the UX better and avoid saving data which we would consider invalid.

The diff here is a bit large, but there shouldn't be anything too complex here, and some of the changes are just due to existing validation tests changing position without any actual behaviour changes.

Another step towards https://trello.com/c/k81APmbi/191-expired-maintenance-should-still-show-in-table-users-can-always-change-requested-start-and-end-date-when-confirming; based on #100.

### DIFF
--- a/app/models/maintenance_window.rb
+++ b/app/models/maintenance_window.rb
@@ -122,8 +122,39 @@ class MaintenanceWindow < ApplicationRecord
 
   def validate_requested_period
     return unless requested_start && requested_end
+    validate_start_before_end
+    validate_start_or_end_in_future_if_needed
+  end
+
+  def validate_start_before_end
     if requested_start > requested_end
       errors.add(:requested_end, 'must be after start')
+    end
+  end
+
+  def validate_start_or_end_in_future_if_needed
+    return if maintenance_period_can_be_passed?
+    validate_field_in_future(:requested_start)
+    validate_field_in_future(:requested_end)
+  end
+
+  def maintenance_period_can_be_passed?
+    # In any of these states the maintenance period can be passed, otherwise
+    # existing saved MaintenanceWindows would become invalidated.
+    #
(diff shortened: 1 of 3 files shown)


WilliamMcCumstie:
Rebased on develop

app/models/maintenance_window.rb
```diff
@@ -119,6 +120,44 @@ def number_associated_models
     [cluster, component, service].select(&:present?).length
   end
 
+  def validate_requested_period
+    return unless requested_start && requested_end
```
WilliamMcCumstie: Is it possible for the requested start/end not to be set? If so what happens if only one of them is set?

Surely there should still be some validation on the single date that is set.

WilliamMcCumstie:
I'll merge this through now, my comment can be addressed later.

app/models/maintenance_window.rb
bobwhitelock: If either (or both) of these are unset then the model will already be invalid as we validate their presence:
https://github.com/alces-software/alces-flight-center/blob/361051bee2bd8e321e2d9ef6f26bc28099c34ca6/app/models/maintenance_window.rb#L10-L11
So I think there is no need to handle this situation in other validations apart from by making sure they do not error, which this line does, as we will need to fix these fields not being present anyway at which point this validation will run.

PR #104, opened 2018-03-14
MaintenanceWindow-related test tweaks

bobwhitelock:
Nothing major here, just:

- making things clearer/DRYer in `maintenance_window_spec`;

- improving test coverage in `maintenance_window_decorator_spec`.

PR #105, opened 2018-03-14
Support transitioning expired MaintenanceWindows

bobwhitelock:
This PR adds model and Cluster view support for `expired` MaintenanceWindows being able to be transitioned out of this state, including:

- it is now valid for `expired` MWs to transition out of this state, similar to `requester` MWs;

- `expired` MWs are now shown in the maintenance table, with an indication that they are expired and what this means.

Based on #104.


WilliamMcCumstie:
Rebased on develop

PR #107, opened 2018-03-14
Implement case archiving UI as an index page

WilliamMcCumstie:
Based on #93, please merge it first.

This PR implements the `UI` aspect for archiving cases. To facilitate this, the `cases table` has been split into a `index` page that can be integrated into the `cluster` tab bar. When viewing the index, the archive is off by default. However sending a `archive=true` parameter with the request will switch it on.

The buttons for adding a new case/ consultancy have been merged into the tab bar. However watch this space as that maybe reverted back in the future.

A spec has also been added for the creation of the `nav_links`. This allowed for the code to be cleaned up whilst ensuring the functionality is the same.

This PR may cause a few issues with the `UI` on non-tab pages, however this will be fixed when tabs are used across the board.

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -12,6 +12,7 @@ class CasesController < ApplicationController
              end
 
     current_site.cases.map(&:update_ticket_status!) if current_user.admin?
+    @archive = true
   end
 
   def show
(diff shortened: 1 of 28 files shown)


bobwhitelock:
Mostly LGTM, thanks; I've just made a couple of tweaks and will merge the rebased version (#123) once CI  passes.

PR #109, opened 2018-03-15
Feature/refactor tabs rendering

WilliamMcCumstie:
Based on #107, please merge it first.

This implements the changes recommend in https://github.com/alces-software/alces-flight-center/pull/92#discussion_r172597002.

The `cluster/tabs` partial has been renamed `partials/tabs`. All the cluster specific code has been moved into a decorator method `tabs` which returns the array used to render the tab bar. A base `tabs` has been added to ApplicationDecorator so an empty tab bar can be rendered for all models.

PR #110, opened 2018-03-16
Switch the `ComponentGroups` pages to use tabs

WilliamMcCumstie:
Based on #109, please merge it first.

This PR is similar to the work done on the `Cluster` pages. It converts all the `ComponentGroup` pages to use `tabs` instead of one big long page.

The only two pages for `ComponentGroups` are the `asset_record` and `components`. A new route had to be added for both of these pages.

This PR does screw around with the `UI` for asset records. This will be fixed once the `components` have been switched to using tabs as well. Until then, there will be some double rendering of header bars. Also the `edit` button has disappeared for reasons. This will be fixed with the other `UI` issues. See: https://trello.com/c/Y26b2QCo/72-clean-up-assetrecordstable-partial

To help share code between the different tab bars, a new `TabsHelper::TabsBuilder` object has been created. It allows the code for generating tabs to be reused on multiple sites. The `TabsBuilder` can be accessed in all the decorators via the `tabs_builder` method. This handles the initialisation of object in the background.

The `UI` concerning the headers is also messed up. This is an ongoing issue with the switching to tabs which will be addressed at the end.

Currently there isn't much code reuse of tabs as only the `Clusters` and `ComponentGroups` pages uses them. This will change once `Components` and `Services` are switched over.


bobwhitelock:
LGTM, was merged in https://github.com/alces-software/alces-flight-center/pull/125.

PR #111, opened 2018-03-17
Implement tabs for the components pages

WilliamMcCumstie:
Based on #110, please merge it first

This is a smallish PR switching the `Components` pages to using the tabs. I won't go into detail here about it as it uses the same design pattern as the other pages. See commits for details.


bobwhitelock:
LGTM, was merged in https://github.com/alces-software/alces-flight-center/pull/126.

PR #136, opened 2018-03-26
Add maintenance durations

bobwhitelock:
This PR is the first step towards changing the time slot for a MaintenanceWindow from being specified as a `requested_start` and `requested_end` to being specified by a `requested_start` and a `duration` (in days). The maintenance should then last from `requested_start` until `duration` business days later, with only the former value being changeable by Site contacts when they confirm maintenance (so we always have the number of days we request to carry out maintenance).

This PR consists of most steps to remove/update our usage of `requested_end` and replace it with `duration` instead; updating the actual maintenance form to reflect this change is still to come.

Trello: https://trello.com/c/74FbrsaD/200-change-maintenance-form-to-show-requested-start-and-duration.

### DIFF
--- a/app/controllers/maintenance_windows_controller.rb
+++ b/app/controllers/maintenance_windows_controller.rb
@@ -69,7 +69,8 @@ class MaintenanceWindowsController < ApplicationController
   ].freeze
 
   def request_maintenance_window_params
-    params.require(:maintenance_window).permit(REQUEST_PARAM_NAMES)
+    # XXX Get duration from request.
+    params.require(:maintenance_window).permit(REQUEST_PARAM_NAMES).merge(duration: 1)
   end
 
   def confirm_maintenance_window_params
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,7 @@ gem 'bootsnap', require: false
 gem "sentry-raven"
 gem 'state_machines-activerecord'
 gem 'state_machines-audit_trail'
+gem 'business_time'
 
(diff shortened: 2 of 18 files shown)


app/controllers/maintenance_windows_controller.rb
```diff
@@ -69,7 +69,8 @@ def cancel
   ].freeze
 
   def request_maintenance_window_params
-    params.require(:maintenance_window).permit(REQUEST_PARAM_NAMES)
+    # XXX Get duration from request.
```
WilliamMcCumstie: You mention that we set the duration and the client can only change the start date. If this is the case, I don't think we should be getting the duration from the request.

Instead, it can be retrieved from the `db`

bobwhitelock: Not sure what you mean here? By 'request' I mean we should get the duration from the `params` when the admin `POST`s to the `request` action - this needs to be done so the duration can be set at all, rather than just hard-coding it to `1` as is done in this PR. This has already been done in the PR based off this one. User's will not be able to change the `duration` as `duration` will not be included in the whitelisted params for the `confirm` action (and they do not have access to `request`), so even if a user manually crafted `POST` data and added a `duration` it would be ignored by `confirm`.

WilliamMcCumstie: Okay, so it is to do with the initial creation of the `maintenance` rather than the update. I figured the next PR probably set it to something other than :)

PR #138, opened 2018-03-26
Standardize the (sub)titles across the pages

WilliamMcCumstie:
This PR address the following cards:
https://trello.com/c/BTiEUnDM/212-make-title-and-subtitle-properties-of-views-rather-than-controllers and
https://trello.com/c/Gg2o45CJ/208-fix-case-archive-link-to-work

The titles where inconsistent because they where being individually set within all the controllers. Instead the `title` is now being set within the `ApplicationController`. The title for all pages is `Management Dashboard - #{scope.name}`. The only exception to this is the root page for `admins` as they do not have a scope. In this case, the title is still being manually set.

The `subtitle` no longer uses an instance variable set within a controller. Instead, each view uses `content_for` to set the subtitle. This allows the same view to be rendered in multiple places, whilst guaranteeing that the subtitle is set correctly.

The bug where the cases archive link went to the regular cases page has also been fixed https://github.com/alces-software/alces-flight-center/commit/1b77314b78e4c8069b010f8f6f90cf61b3c76ae2.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,7 +7,8 @@ class ApplicationController < ActionController::Base
 
   before_action :set_sentry_raven_context
   before_action :assign_current_user
-  before_action :define_navigation_variables
+  before_action :assign_scope
+  before_action :assign_title
 
   rescue_from ReadPermissionsError, with: :not_found
 
@@ -42,7 +43,7 @@ class ApplicationController < ActionController::Base
     params[id_method] || params[:id]
   end
 
-  def define_navigation_variables
+  def assign_scope
     return unless current_user
 
     @scope = case request.path
@@ -66,6 +67,10 @@ class ApplicationController < ActionController::Base
                          current_user.site
                        end
              end
+  end
+
+  def assign_title
+    @title = "Management Dashboard - #{@scope.name}" if @scope
   end
 
   def format_errors(model)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -70,7 +70,12 @@ class ApplicationController < ActionController::Base
   end
 
   def assign_title
-    @title = "Management Dashboard - #{@scope.name}" if @scope
+    if @scope
+      @title = <<~EOF.squish
+        #{@scope.readable_model_name.split.map(&:capitalize).join(' ')}
+        Dashboard - #{@scope.name}
+      EOF
+    end
   end
 
   def format_errors(model)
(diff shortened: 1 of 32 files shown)


app/views/layouts/application.html.erb
```diff
@@ -34,8 +34,9 @@
               <div class="col"></div>
               <div class="col-8">
                 <h1 class="display-4"><%= @title %></h1>
-                <% if @subtitle %>
-                  <p class="lead"><%= @subtitle %></p>
+                <% subtitle = yield :subtitle %>
+                <% if subtitle.present? %>
+                  <p class="lead"><%= subtitle %></p>
```
bobwhitelock: Not sure we need to make this conditional anymore - if `content_for :subtitle` isn't provided then `yield subtitle` will just be `nil`, so we could simplify this bit slightly to just `<p class="lead"><%= yield :subtitle  %></p>`.

WilliamMcCumstie: So the `yield :subtitle` returns an empty string if the subtitle isn't provided. However it will still render the `<p class='lead'>` tags. This has implications for the white-space as the `<p>` tag will still have the same padding.

bobwhitelock: Yep this is true, I tried this out though and thought it was slightly better as it means the page jumps a bit less when navigating between pages with/without a subtitle, as the title bar changes size less.

app/controllers/clusters_controller.rb
```diff
@@ -1,16 +1,5 @@
...
-
-    support_type = case @cluster.support_type.to_sym
-                   when :managed
-                     'Managed'
-                   when :advice
-                     'Self-managed'
-                   end
-    @subtitle = "#{support_type} cluster"
-  end
+  def show; end
```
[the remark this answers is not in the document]
WilliamMcCumstie: Removed

app/views/layouts/application.html.erb
WilliamMcCumstie: Updated

app/views/components/index.html.erb
```diff
@@ -1,4 +1,4 @@
-
+<% content_for :subtitle { 'Components' } %>
```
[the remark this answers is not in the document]
WilliamMcCumstie: Done

app/controllers/application_controller.rb
```diff
@@ -68,6 +69,10 @@ def define_navigation_variables
              end
   end
 
+  def assign_title
+    @title = "Management Dashboard - #{@scope.name}" if @scope
```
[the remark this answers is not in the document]
WilliamMcCumstie: Done

WilliamMcCumstie:
I've addressed your comments and it should hopefully be ready to merge. I'll look at fixing up the `sites` page in the next PR

bobwhitelock:
Looks great, thanks :+1:

PR #139, opened 2018-03-27
Compulsory maintenance

bobwhitelock:
This PR implements the idea of compulsory or mandatory maintenance, which bypasses the usual maintenance confirmation process (requiring a site contact to confirm and possibly reschedule all maintenance). Briefly how this works is:

- there is a new `mandate` event which transitions a maintenance window directly from `new` to `confirmed`; this therefore bypasses the usual maintenance confirmation process and so an appropriate comment, distinct from the usual `confirm` or `request` comment, is left on the RT ticket indicating this;

- a new check box is included in the request maintenance form, when checked this means the new maintenance window will be `mandate`d rather than `request`ed and so bypasses the usual process as described above.

Some minor related refactoring is also included here.

Trello: https://trello.com/c/hjAR3loz/192-compulsory-maintenance.

Based on #137.


WilliamMcCumstie:
Rebased on develop

app/controllers/maintenance_windows_controller.rb
```diff
@@ -8,11 +8,13 @@ def new
   end
 
   def create
-    handle_form_submission(action: :request, template: :new) do
+    event, action =
+      params[:mandatory] ? [:mandate, :schedule] : [:request] * 2
```
WilliamMcCumstie: This line is quite dense and it took me a moment to work out what it was doing. Considering it is all ready split between two lines, I think it would be clearer as two separate ternary statements. A `mandatory?` method could also be useful.

```
event = mandatory? ? :mandate : :request
action = mandatory? ? :schedule : :request

....

def mandatory?
  !!params[:mandatory]
end
```

PR #140, opened 2018-03-27
Split the site page out into tabs

WilliamMcCumstie:
This PR switches the `sites` page to use `tabs`. The cluster over view has been removed from this page as it is duplicates the landing page for the cluster itself. Instead a link has been provided to the cluster.

In order for the tabs to work, the method for finding the paths of the scope had to be updated. This is because the contacts and admins use different paths for the sites. Admins are the same as before, the difference is that contacts only have one site. This means contacts do not visits `/site/1/cases` but `/cases` directly.

Also the additional contacts are not displayed on the page if there is none.


app/helpers/tabs_helper.rb
```diff
@@ -27,11 +27,24 @@ def asset_record
 
     attr_reader :scope, :h
 
+    def page_is_a_contacts_site?
+      scope.is_a?(Site) && h.current_user.contact?
+    end
+
+    # Regular contacts have their site as the root path and thus need to be
+    # handled separately
     def scope_path(method = nil, **hash_params)
```
bobwhitelock: This method seems quite dense and convoluted now. Rather than having all this embedded here we could extract it to the various model decorators, which could implement a `path` method, and then we could then just call `scope.path(method, **hash_params)` and have the appropriate path helper be worked out and used.

This would be more reusable and easily testable than having this all embedded in `TabsBuilder` (which shouldn't really be directly concerned with forming paths). It would also mean we wouldn't need to switch on the type of `scope`, like `scope.is_a?(Site)` (which can often be a clue that we should be using polymorphism instead).

Note that we already have a `path` method in several of the decorators, which currently just turns things like `cluster.path` into `cluster_path(cluster)` so we can treat different models in the same way. This would become just one case in the new method I believe, as calling `scope.path` without arguments should have the same effect as this.

There are some other places we also have logic which looks similar, like these:

https://github.com/alces-software/alces-flight-center/blob/f03b8857bc7ba0c9940e90d0ae511d0b2db5e5db/app/decorators/site_decorator.rb#L8-L12

https://github.com/alces-software/alces-flight-center/blob/f03b8857bc7ba0c9940e90d0ae511d0b2db5e5db/app/decorators/application_decorator.rb#L83-L96

So I think it may be possible to tidy up how we determine dynamic paths in various places to use the same method.

When implementing this, I think it will require similar code to the current method in `ApplicationDecorator`, and then some parts of this will need to be overridden in `SiteDecorator` to do the special handling needed for the site paths.

app/views/sites/show.html.erb
```diff
@@ -23,31 +12,16 @@
...
-        </li>
-      </ul>
-    </div>
-
-    <%= render 'clusters/details', cluster: cluster %>
-  </div>
+  <% site.clusters.each do |cluster| %>
+    <li class='list-group-item list-group-item-action text-center'>
+      <%= link_to "#{cluster.name} - Manage this cluster",
+                  cluster_path(cluster) %>
```
bobwhitelock: This would look better as a large button I think (or link with button role at least, similar to the Case form buttons); it looks slightly odd at the moment as the whole row is highlighted when hovered but only the small link in the centre does anything.

app/views/sites/show.html.erb
```diff
@@ -23,31 +12,16 @@
...
-          %>
-        </li>
-      </ul>
-    </div>
-
-    <%= render 'clusters/details', cluster: cluster %>
-  </div>
+  <% site.clusters.each do |cluster| %>
+    <li class='list-group-item list-group-item-action text-center'>
+      <%= link_to "#{cluster.name} - Manage this cluster",
```
bobwhitelock: This would look better as `"#{cluster.name}: Manage this cluster"` I think. Also a hyphen is often not what you want in this sort of text, [you may want an en-dash instead](https://english.stackexchange.com/questions/2116/when-should-i-use-an-em-dash-an-en-dash-and-a-hyphen).

app/helpers/tabs_helper.rb
```diff
@@ -27,11 +27,24 @@ def asset_record
...
-      method_string = [
-        scope.class.name.underscore, method, 'path'
-      ].reject(&:nil?).join('_')
-      h.send(method_string, scope, **hash_params)
+      method_args = begin
+        if page_is_a_contacts_site?
+          [(method ? method : 'root') + '_path']
+        else
+          method_string = [
+            scope.class.name.underscore, method, 'path'
```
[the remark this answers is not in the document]
WilliamMcCumstie: This has been changed in a future PR

app/helpers/tabs_helper.rb
```diff
@@ -27,11 +27,24 @@ def asset_record
...
+          ].reject(&:nil?).join('_')
```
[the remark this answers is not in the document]
WilliamMcCumstie: The method has been removed in a future PR

app/helpers/tabs_helper.rb
WilliamMcCumstie: This has been addressed in a future PR

app/views/sites/show.html.erb
WilliamMcCumstie: Of course there is a grammar guide on the length of dashes. Anyway, I have fixed it.

app/views/sites/show.html.erb
WilliamMcCumstie: This has been fixed in a later PR

bobwhitelock:
Thanks, will look into the changes made in future PRs when I review them.

PR #141, opened 2018-03-28
Update the case archives to use their own route

WilliamMcCumstie:
Based on #140, please merge it first.

This PR primary focus is to switch the the archive cases to use there own route. Previously a magic parameter had to be sent with the request to the cases index (`/cases?achive=true`). This has been removed and a route to `/cases/archives` has been added in its place. This new route uses a custom action `archives` that wraps the `index` action. This was the easy bit.

The harder part is getting all the links to use this new route. Previously the `TabsBuilder` had its own method for building the path helpers. However the archive link paths now looks like `archives_component_cases_path`. This is quite a bit trickier to build, so the `TabBuilder.scope_path` method has been replaced by a decorator `method_missing`.

Now all methods that contain `scope_***path` can be automatically converted to the corresponding helper method. This means `scope.archives_scope_cases_path` is equivalent to `archives_cluster_cases_path(scope)` if the scope is a cluster. This is done by some fancy regex and substitutions.

It also handles the difference in routing to the sites pages for `admins` and `contacts`. For admins it is business as usual, however for contacts it automatically takes into account it needs to be the path from the root (note `/sites/:id`).

This addresses the following comments (I made most of the changes before the PR was reviewed):
https://github.com/alces-software/alces-flight-center/pull/140#discussion_r177557249
https://github.com/alces-software/alces-flight-center/pull/140#discussion_r177546931
https://github.com/alces-software/alces-flight-center/pull/140#discussion_r177547477


bobwhitelock:

app/decorators/application_decorator.rb
```diff
@@ -57,8 +57,38 @@ def invalid_feedback_div(field_name)
...
+      .sub(/\Apath\Z/, 'root_path')
+      .to_sym
+  end
+
+  # This method is overridden in the Site decorator
+  def scope_name_for_paths
+    '_' + model.underscored_model_name + '_'
+  end
+
+  # This method is overridden in the Site decorator
```
bobwhitelock: Don't need the comments here and above I think - if someone is interested in how things are handled for Sites in particular then they can go look at that code, and otherwise this is the sort of comment which will get out-of-date if various things change (the SiteDecorator or this decorator could change, or if other decorators later override these methods these comments would be misleading).

app/decorators/application_decorator.rb
```diff
@@ -57,8 +57,38 @@ def invalid_feedback_div(field_name)
     )
   end
 
+  def method_missing(s, *a, **hash, &b)
+    if respond_to_missing?(s, *a) == :scope_path
```
bobwhitelock: This seems a nice way to use `respond_to_missing?` from `method_missing` and ensure they always handle the same methods, good idea :+1:.

app/decorators/application_decorator.rb
```diff
@@ -57,8 +57,38 @@ def invalid_feedback_div(field_name)
...
+
+  def respond_to_missing?(s, *_a)
+    s.match?(/\A(.+_)?scope_(.+_)?path\Z/) ? :scope_path : super
+  end
+
   private
 
+  def convert_scope_path(s)
+    (s.match(/\Ascope_.*/) ? "_#{s}" : s.to_s)
+      .sub(/_scope_/, scope_name_for_paths)
```
bobwhitelock: Could these 2 lines be replaced with something like this:

```ruby
s.sub(/_?scope_/, scope_name_for_paths)
```

?

Not certain, but looks like these are ensuring the string contains `_scope_` so we can then replace it, when we could match both `_scope_` and `scope_` in the first place.

app/decorators/application_decorator.rb
WilliamMcCumstie: I have reword the comments. I have removed the reference to the `SiteDecorator` and instead made a mention that the method can be overridden.

app/decorators/application_decorator.rb
WilliamMcCumstie: Thanks. I've been trying to tackle the best way to handle `respond_to_missing?` for a while now.

app/decorators/application_decorator.rb
WilliamMcCumstie: I have simplified down those two lines of code. Not exactly how you mentioned as that could have side affects in the future. As we are using `regex`, it might as well match the pattern exactly as it also documents the intended behaviour. In this case we want to either match `_scope_` OR methods that start with `scope_` exactly.

This can be achieved by using capture groups which are included in the substitution. The `scope_name_for_paths` has also been adjusted so it does not need to include the leading and trailing `_`.

PR #142, opened 2018-03-28
Improve MaintenanceWindow transition handling

bobwhitelock:
This PR contains several tweaks to how MaintenanceWindow events and transitions are handled. These don't/shouldn't affect anything from a user's perspective, but should make things more straightforward for implementing https://trello.com/c/RtUVccsL/201-add-ability-to-end-started-maintenance-early. In particular this includes:

- changing whether we validate whether a transition should have a user present/absent to be based on the transition's `event`, rather than the state it is transitioning `to` - this is better as we need to have multiple ways for transitioning between states, both automatic and user-initiated;

- renaming the automatic events triggered when we automatically process MaintenanceWindows to change their state based on the current time, if needed,  so that these are prefixed with `auto_` - this is clearer and means we can have multiple similar events prefixed this way or not depending on how they are triggered.

Based on #139.


WilliamMcCumstie:
Rebased on develop

PR #144, opened 2018-03-29
Ending maintenance early

bobwhitelock:
This PR adds support for admins to end maintenance early via a button on the Cluster maintenance page; this functions similarly to maintenance being automatically ended when its `expected_end` time is reached, but allows them to return components etc. to service early if needed and leaves a comment on the RT ticket appropriate to this.

Based on #142.
Trello: https://trello.com/c/RtUVccsL/201-add-ability-to-end-started-maintenance-early.


WilliamMcCumstie:
Rebased on develop

PR #143, opened 2018-03-29
Various small tweaks

WilliamMcCumstie:
Based on #141, please merge it first.

This PR contains various small tweaks as suggested in previous PRs and cards. Please see commits for more details.

PR #145, opened 2018-04-03
Clean up the asset_record and component_expansions tables

WilliamMcCumstie:
Based on #143, please merge it first. There is currently a conflict with develop, which will need to resolved once the other PRs are merged. Likely this branch will need to be rebased.

The PR addresses the outstanding issues in the following cards. Some of which has already been addressed.
https://trello.com/c/Y26b2QCo/195-clean-up-assetrecordstable-partial
https://trello.com/c/owhM2PME/196-clean-up-expansionstable

The `asset_record_table` and `expansions_table` were implemented before the `scope` was fully rationalised. They now use the instance variables that where are set in the `ApplicationController`. In the case of the expansions, the `component` instance variable is used as only components can have expansions.

Also both of the partials have been moved out of the `partials` directory as they are specific to their respective models. Instead they have been placed closer to the templates that use them.


bobwhitelock:

PR #146, opened 2018-04-03
Hide archived cases from select form

WilliamMcCumstie:
This address card: https://trello.com/c/3E1HFmrM/170-hide-archived-cases-from-any-drop-down-when-asked-to-choose-tickets

From now on, archived cases will not appear in the select box. This is done with the use of a partial to generate the input. It is used in both the `logs` and `maintenance` pages. The partial supports selecting both a single and multiple cases.

The spec for the maintenance window had to be adjusted as it used `data-tags` to assist with the `Capybara` selection. These tags do not generalise well and have been removed in this instance. Instead the `id` for the specific field is used instead.

A rebase was required, so the commits might references changes that haven't happened.


bobwhitelock:
Rebased on `develop` for easier reviewing and to remove duplicated commits.

PR #152, opened 2018-04-04
Use Flight SSO for Flight Center

jamesremuscat:
This PR changes Flight Center to use Flight SSO for authentication.

This affects a couple of the existing workflows and processes:

- The password-changing workflow no longer applies. Users can reset their own passwords via the Flight SSO workflow.
- Center accounts are tied to Flight accounts by email address. To add a user to Center, a Center account must be manually created (using the existing process) with an email matching a user's Flight account.

Users signed in to Flight without a matching Center account are treated as logged out. There are some words explaining the situation in `app/views/sessions/new.html.erb` which we might want to tweak.

AFAICT the only reason this won't merge automatically or rebase easily is that `flight-chrome.js` has been changed a number of times both here and on `develop`. We'll ultimately want the most recent version of that file at the time we merge (or merge and then update `flight-chrome`, whichever is easiest, I suppose).

In order to not block releases of `flight-center` and `flight-chrome` between now and when we deploy these changes live, there's a global variable that controls whether or not the Flight account menu is displayed on Flight Center's brand bar. By default this will mean the menu is hidden (existing behaviour). This branch includes the change needed to make that menu visible. Once the changes on this branch hit live we can eliminate the global variable from `flight-chrome` and here in a future release.

Trello: https://trello.com/c/gUKttsif/194-implement-support-for-flight-sso


jamesremuscat:
I should also probably note that I haven't tried to unpick the fields Clearance added to the User model (e.g. `encrypted_password`, `remember_token`, so they're still there but now unused. It's possible we could remove them but it might cause Clearance to become upset - I haven't verified this.

app/controllers/application_controller.rb
```diff
@@ -19,6 +20,11 @@ def error_flash_models(models, header)
     end.join("\n")
   end
 
+  def sign_out
+    clearance_session.sign_out
+    cookies.delete('flight_sso', domain: request.host[request.host.index('alces')..-1])
```
bobwhitelock: Is the `request.host.index('alces')..-1` to get `'alces-flight.com'`? Do we want to avoid using this directly for some reason?

app/models/user.rb
```diff
@@ -43,4 +43,22 @@ def info
...
+
+    user.tap do |u|
+      # If we want to update local values with those given in the token, do the following:
+      #  u.name = claims.fetch('username')
+      #  u.save
+    end
+  end
+
+  def remember_token
+    # This is only here for testing purposes
```
bobwhitelock: I don't quite follow this - will this not always be used?

app/views/sessions/new.html.erb
```diff
@@ -36,7 +36,12 @@
       </div>
```
bobwhitelock: This card now still shows 'Sign in' as a title but no longer contains a way to sign in which may confuse Flight users when they first visit Flight Center without an account - possibly the wording could be improved slightly? Not sure exactly of what words would be best, @mjtko may have opinions :slightly_smiling_face:.

config/environments/development.rb
```diff
@@ -60,4 +60,6 @@
 
   # Alces Flight Center custom config.
   config.rt_interface_class = FakeRequestTrackerInterface.to_s
+
+  config.sso_base_url = ENV['SSO_BASE_URL'] || 'http://accounts.alces-flight.lvh.me:4000'
```
bobwhitelock: To use the SSO for Flight Center in development is it required to be running flight-sso somewhere locally, by default on port 4000? Are there any external docs related to using this we could point to from this repo, or can we add some brief documentation on this to this repo? I've not used flight-sso before so not sure how I should be doing this.

Probably the README would be the obvious place to link/add this, and/or we have some steps for getting started output at the end of `bin/setup`, so that would also be a good place to mention what else will be needed to log in.

app/models/user.rb
jamesremuscat: We're overriding Clearance's default behaviour so that it never writes a `remember_token` in anything other than tests, so it no longer calls this method. (The `remember_token` is now actually the `flight_sso` cookie, so it's the SSO server that writes it, and Clearance/Center only reads it.)

Some of the tests depend on persisting user auth across requests, so they need to be able to write a (sparse) SSO-like cookie, and that's what uses this method.

config/environments/development.rb
jamesremuscat: I've written some words in the README plus added a pointer to the end of `bin/setup`.

app/views/sessions/new.html.erb
jamesremuscat: Sadly it would be non-trivial (but possible) to add a "log in" button or form to this box. Happy to take input on the wording!

bobwhitelock:
Thanks for your explanations and making those updates :slightly_smiling_face:.

Just had a look at this again and tested out actually using an SSO account and signing in. This all looks good to me I think, there's just a few things we should remember to do before deploying this which I've added to the corresponding card.

Only other minor comment I have is that it might be useful to inline your explanations above as code comments, to make them obvious in future.

jamesremuscat:
Rebased on latest `develop` and fixed many conflicts.

bobwhitelock:
As well as the above issue, which is needed to release this, we also need to look into https://trello.com/c/7Ml8EKQ9/333-investigate-what-if-any-changes-are-needed-to-use-flight-sso-support-in-staging.

bobwhitelock:
Rebased again on latest `develop`.

bobwhitelock:
@jamesremuscat @benarmston Thanks for you work on this both, I think we're ready to merge this now. I've just made a couple of additions, in particular making it so when we import the production backup for development we also create corresponding SSO `Account`s for Flight Center `User`s (where these do not already exist). I've also merged together the SSO docs with the development setup steps in the README, so this should now provide all the steps needed to set up a dev environment in one place.

PR #154, opened 2018-04-05
Automatically archive completed cases after two weeks

jamesremuscat:
This PR adds the ability to automatically archive cases whose corresponding RT tickets have been completed at least two weeks ago.

An hourly rake cron task updates ticket statuses for all non-archived `Case`s.

If a ticket has transitioned from a non-completed state to a completed one (or fsr a resolution time wasn't previously recorded), the RT resolution time is recorded (if present - else current time is used) as the `Case`'s "completion time".

If the completion time is two weeks ago or greater, then the `archived` flag is also set on that `Case` and this change is logged to a file.

Trello: https://trello.com/c/OetM2i5j/169-auto-archive-cases-when-resolved-for-2-weeks

### DIFF
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -76,7 +76,7 @@ class Case < ApplicationRecord
   end
 
   def update_ticket_status!
-    return if ticket_completed? && !self.completed_at.nil?
+    return if ticket_completed? && self.completed_at
     self.last_known_ticket_status = associated_rt_ticket.status
     if ticket_completed?
       self.completed_at = associated_rt_ticket.resolved || DateTime.now.utc
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -79,7 +79,7 @@ class Case < ApplicationRecord
     return if ticket_completed? && !self.completed_at.nil?
     self.last_known_ticket_status = associated_rt_ticket.status
     if ticket_completed?
-      self.completed_at = associated_rt_ticket.resolved
+      self.completed_at = associated_rt_ticket.resolved || DateTime.now.utc
     end
     save!
   end
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -105,6 +105,10 @@ class Case < ApplicationRecord
     associated_model.readable_model_name
   end
 
+  def ticket_completed?
+    COMPLETED_TICKET_STATUSES.include?(last_known_ticket_status)
+  end
+
   private
 
   def create_rt_ticket
@@ -128,10 +132,6 @@ class Case < ApplicationRecord
 
   def assign_default_subject_if_unset
     self.subject ||= issue.default_subject
-  end
-
-  def ticket_completed?
-    COMPLETED_TICKET_STATUSES.include?(last_known_ticket_status)
   end
 
   def associated_rt_ticket
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -135,7 +135,7 @@ class Case < ApplicationRecord
   end
 
   def associated_rt_ticket
-    rt.show_ticket(rt_ticket_id)
+    @associated_ticket ||= rt.show_ticket(rt_ticket_id)
   end
 
   def rt
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -79,7 +79,7 @@ class Case < ApplicationRecord
     return if ticket_completed? && self.completed_at
     self.last_known_ticket_status = associated_rt_ticket.status
     if ticket_completed?
-      self.completed_at = associated_rt_ticket.resolved || DateTime.now.utc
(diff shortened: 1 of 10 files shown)


app/models/case.rb
```diff
@@ -76,9 +76,11 @@ def open
   end
 
   def update_ticket_status!
-    return if ticket_completed?
-
+    return if ticket_completed? && !self.completed_at.nil?
     self.last_known_ticket_status = associated_rt_ticket.status
+    if ticket_completed?
+      self.completed_at = associated_rt_ticket.resolved || DateTime.now.utc
```
bobwhitelock: Have you checked if the `self.completed_at = associated_rt_ticket.resolved` works when using a real ticket retrieved from RT? I think it will work in tests as we're explicitly setting `resolved` to `Time.now` (here: https://github.com/alces-software/alces-flight-center/pull/154/files#diff-16eb08ba34f4646e1499abd12aa77583R19), but when retrieving real tickets from RT the `resolved` field seems to be a string in an odd format:

```ruby
[18] pry(main)> rt = RequestTrackerInterface.new
=> #<RequestTrackerInterface:0x0055b006278510 @api_endpoint="http://helpdesk.alces-software.com/rt/REST/1.0/", @password="ruhSBVLXd5tHDfVzbpFI", @username="bob.whitelock">
[19] pry(main)> rt.show_ticket(1)
=> #<struct
 id=1,
 queue="Support",
 owner="steve.norledge",
 creator="steve@alces-software.com",
 subject="test",
 status="deleted",
 priority="0",
 initialpriority="0",
 finalpriority="0",
 requestors="steve@alces-software.com",
 cc="",
 admincc="",
 created="Sat Nov 14 18",
 starts="Not set",
 started="Sat Nov 14 18",
 due="Not set",
 resolved="Mon Aug 23 21",
 told="Sat Nov 14 19",
 lastupdated="Mon Aug 23 21",
 timeestimated="0",
 timeworked="0",
 timeleft="0",
 cf_supportyear="",
 cf_credits="",
 cf_category="",
 cf_clusterid="">
```

I'm not sure what format these dates are supposed to be, as the same ticket in RT's own interface was resolved in 2010: http://helpdesk.alces-software.com/rt/Ticket/Display.html?id=1 (rather than 2021 :thinking:).

I think it's was a good idea to try to set `completed_at` to the `resolved` time from the RT ticket, rather than just what the time is when we retrieve it, and if we planned to keep RT integration long term it would be good to figure out what format they're using and parse this. Since the immediate plan is now to migrate away from integrating with RT though,  and move Case management fully into Flight Center, I think we can remove the `associated_rt_ticket.resolved ||` and just set `completed_at` to the current time.

app/models/case.rb
```diff
@@ -76,9 +76,11 @@ def open
   end
 
   def update_ticket_status!
-    return if ticket_completed?
-
+    return if ticket_completed? && !self.completed_at.nil?
```
bobwhitelock: Here and in `cases.rake` we check `&& !completed_at.nil?` - I believe we could achieve the same thing more concisely using `&& !completed_at` directly.

lib/tasks/cases.rake
```diff
@@ -0,0 +1,29 @@
...
+      logger.info("Automatically archiving cases completed on or before #{2.weeks.ago}")
+
+      Case.where(archived: false).each do |c|
+        c.update_ticket_status!
+
+        if c.ticket_completed? && !c.completed_at.nil? && c.completed_at <= 2.weeks.ago
+          c.archived = true
+          logger.info("Archiving case #{c.id} completed at #{c.completed_at}")
+        end
+        c.save!
```
bobwhitelock: I think we could use `c.update!(archived: true)` inside the `if` here to indicate that only those Cases being archived need saving and to avoid saving every unarchived Case, which isn't needed.

app/models/case.rb
jamesremuscat: `RequestTrackerInterface#ticket_struct_from_data` was getting confused by the colons in the date string - it's far more sane than that snippet makes it out to be!

```
curl -b RT_SID_helpdesk.alces-software.com.80=XXXXX http://helpdesk.alces-software.com/rt/REST/1.0/ticket/922/show
RT/3.6.7 200 Ok

id: ticket/922
[...blah]
Resolved: Mon Nov 29 10:23:05 2010
```
In any case, now changed to just use the current time.

bobwhitelock: Ah, that's much more reasonable than it looked like to me - I should know better than to trust code I wrote several months ago!

bobwhitelock:
Looks good, can I merge now or is there more to come?

jamesremuscat:
Merge away :-)

PR #155, opened 2018-04-10
Add comments to cases

jamesremuscat:
This PR adds the ability for users to add comments to cases.

A user may comment on any `Case` within their `Site`. Admins may comment on any `Case`. Comments are displayed from oldest to newest on a `Case`'s page, with the name of the user who posted the comment and the date and time at which they did so.

Empty comments are prohibited, and comments must be at least two characters long ("OK" probably being the tersest possible comment still containing useful information).

Trello: https://trello.com/c/JHENp9CD/217-add-comments-to-cases-2-days-moreish

### DIFF
--- a/app/controllers/case_comments_controller.rb
+++ b/app/controllers/case_comments_controller.rb
@@ -2,12 +2,16 @@ class CaseCommentsController < ApplicationController
   before_action :require_login
 
   def create
-    case_id = params[:id]
+    case_id = params[:case_id]
     my_case = Case.find(case_id)
 
     if my_case.site == current_user.site || current_user.admin?
-      my_case.case_comments.create(user: current_user, text: params[:text])
-      flash[:notice] = 'New comment added.'
+      new_comment = my_case.case_comments.create(user: current_user, text: params[:case_comment][:text])
+      if new_comment.persisted?
+        flash[:notice] = 'New comment added.'
+      else
+        flash[:error] = 'Your comment was not added.'
+      end
     else
       flash[:error] = 'You do not have permission to comment on this case.'
     end
--- a/app/controllers/case_comments_controller.rb
+++ b/app/controllers/case_comments_controller.rb
@@ -5,15 +5,11 @@ class CaseCommentsController < ApplicationController
     case_id = params[:case_id]
     my_case = Case.find(case_id)
 
-    if my_case.site == current_user.site || current_user.admin?
-      new_comment = my_case.case_comments.create(user: current_user, text: params[:case_comment][:text])
-      if new_comment.persisted?
-        flash[:notice] = 'New comment added.'
-      else
-        flash[:error] = 'Your comment was not added.'
-      end
+    new_comment = my_case.case_comments.create(user: current_user, text: params[:case_comment][:text])
+    if new_comment.persisted?
+      flash[:notice] = 'New comment added.'
     else
-      flash[:error] = 'You do not have permission to comment on this case.'
+      flash[:error] = "Your comment was not added. #{new_comment.errors.full_messages.join('; ').strip}"
     end
 
     redirect_to case_path(case_id)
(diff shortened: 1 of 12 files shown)


bobwhitelock:
Rebased on `develop` to resolve conflict.

app/controllers/case_comments_controller.rb
```diff
@@ -0,0 +1,17 @@
+class CaseCommentsController < ApplicationController
+  before_action :require_login
```
bobwhitelock: Don't think we need this guard. There's no harm in it afaics, and looks like we do the same in the `CasesController` (which also is unneeded I believe), but just thought I'd point out this is unneeded as we're doing access control at the routing level so don't need to do this in individual controllers (see https://github.com/thoughtbot/clearance#access-control).

app/controllers/case_comments_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def create
+    case_id = params[:case_id]
+    my_case = Case.find(case_id)
+
+    new_comment = my_case.case_comments.create(user: current_user, text: params[:case_comment][:text])
```
bobwhitelock: Bit of an edge case, but worth pointing out that the `params[:case_comment][:text]` will blow up I believe if we receive POST data in a format other than what we expect. Using `params.require(...).permit(...)` may be desirable to handle this better, in particular if we later want to include more data when commenting.

app/views/case_comments/_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= form_for [@case, @comment] do |f| %>
```
bobwhitelock: TIL you can pass an array to `form_for` to access namespaced routes :slightly_smiling_face:

spec/models/case_comment_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      other_site = create(:site)
+      other_user = create(:user, site: other_site)
+
+      expect do
+        CaseComment.create!(
+           case: @case,
+           user: other_user,
+           text: 'This comment is not allowed'
+        )
+      end.to raise_error(ActiveRecord::RecordInvalid)
```
bobwhitelock: Worth pointing out that in both this and the above case we check that the record is invalid, and so not created, but we're not actually checking that the reason it's invalid is what we expect, e.g. by checking `comment.errors.messages` includes the error we expect. So it would be possible for these tests to give a false positive if the created record was invalid for any other reason; from reading this it doesn't look like this is currently the case, but it could later become so.

spec/factories.rb
```diff
@@ -1,5 +1,12 @@
 
 FactoryBot.define do
+
+  factory :case_comment do
+    user create(:user)
```
bobwhitelock: To just specify an arbitrary relationship in a factory you only need the name of that relationship, i.e. for this line just `user` would be sufficient I believe.

In this case though I believe we might actually want to constrain the particular user which is associated - since we have the constraint that the commenting user must be an admin or a site contact for the case's site, we should specify one of these in this relation otherwise the default comment created by this factory will be invalid; using `create(:admin)` would probably be the simplest way to do this.

As it happens we're not currently using this factory anywhere so not specifying this didn't cause anything to fail, so we may instead want to remove or use this factory somewhere.

app/controllers/case_comments_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    if new_comment.persisted?
+      flash[:notice] = 'New comment added.'
+    else
+      flash[:error] = "Your comment was not added. #{new_comment.errors.full_messages.join('; ').strip}"
+    end
```
bobwhitelock: A couple of feature tests that both these success and failure paths for commenting work as intended, and continue working, would be useful. An example of a similar test is [here](https://github.com/alces-software/alces-flight-center/blob/98fd934b2563b7d94621095bacc1329106ed8714/spec/features/maintenance_windows_spec.rb#L295,L318), where we essentially visit a page, fill in a form, and then assert that both the model and view are changed as expected. Also see [Capybara](https://github.com/teamcapybara/capybara#the-dsl) for details of the API we're using in feature tests, if you're unfamiliar with this :slightly_smiling_face:.

PR #163, opened 2018-04-13
Replace RT with Flight Center-sent emails

jamesremuscat:
This PR builds on the work in the `feature/comment-on-cases` branch (#155), to replace RT as the source of emails sent by Flight Center. (It therefore includes the commits from that branch.)

As a consequence of this PR, tickets are no longer created in RT when cases are raised in FC. Instead, Flight Center itself generates an email with standard Flight formatting, with the same content that would have been previously sent via RT.

We retain the `rt_ticket_id` and `last_known_ticket_status` fields of `Case`, but they are no longer mandatory (and are now nullable in the database). For tickets with an `rt_ticket_id` we maintain the same E-mail subject as before, in the hope that E-mail clients will correctly thread the email chain. We also maintain the existing behaviour of updating ticket status from RT for all appropriate tickets.

Emails currently get sent:
- When a case is created: to all contacts for the site *except* the user creating the case
- When a comment is added to a case: to all contacts for the site *except* the user making the comment (**I believe this is a change to RT behaviour**)
- When a maintenance window state transition happens: to all contacts for the site, using the same text as before (generated by `MaintenanceNotifier`)

Note that all the RT interface code remains present, since we (for now!) still want to track the status of existing RT tickets in FC. Updates to tickets made in FC will no longer be propagated to RT.

Trello: https://trello.com/c/PvkTSpOO/223-replace-creating-rt-tickets-adding-rt-correspondence-with-emailing-users-directly


bobwhitelock:
Rebased on `develop` to resolve conflicts and remove already merged commits.

spec/models/case_spec.rb
```diff
@@ -87,134 +87,6 @@
...
-          hash_excluding(
-            text: /Associated service:/
-          )
-        ).and_return(fake_rt_ticket)
-
-        subject.save!
-      end
-    end
-  end
-
```
bobwhitelock: These tests are all still somewhat relevant I believe, they just need adapting - we no longer need to test that an RT ticket is created correctly, but we should still ensure that an email is generated with the correct text, that this text is altered correctly depending on the properties of the Case, that a correct token is generated and saved for later use etc.

app/models/case_comment.rb
```diff
@@ -11,9 +11,15 @@ class CaseComment < ApplicationRecord
 
   validate :valid_user, on: :create
 
+  after_create :send_comment_email
```
bobwhitelock: A test that commenting on a Case sends an email correctly would be useful. Could either be done in CaseComment model tests, similar to is already done for MaintenanceWindows, or in a feature test testing the full comment process (as suggested in https://github.com/alces-software/alces-flight-center/pull/155/files#r181694427).

app/mailers/case_mailer.rb
```diff
@@ -0,0 +1,30 @@
...
+      cc: @case.email_recipients.reject { |contact| contact == @case.user.email }, # Exclude the user raising the case
+      subject: @case.email_subject
+    )
+  end
+
+  def comment(comment)
+    @comment = comment
+    @case = @comment.case
+    mail(
+      cc: @case.email_recipients.reject { |contact| contact == @comment.user.email }, # Exclude the user making the comment
```
bobwhitelock: This logic is essentially the same as in `new_case`, and future Case-related emails will likely also need similar logic. Rather than doing this here it might therefore make sense to move this into `Case#email_recipients`.

app/models/case.rb
```diff
@@ -121,21 +118,46 @@ def ticket_completed?
...
+    # previous emails related to that Case (see
+    # https://github.com/alces-software/alces-flight-center/issues/37#issuecomment-358948462
+    # for an explanation). If we want to change this format and avoid this
+    # consequence then a solution would be to first add a new field for this
+    # whole string, and save and use the existing format for existing Cases.
+    "#{cluster.name}: #{subject} [#{token}]"
+  end
 
-    self.rt_ticket_id = ticket.id
+  def rt_ticket_text
```
bobwhitelock: This method is now being used in non-RT related contexts so should probably be renamed - possibly `formatted_text`?

app/mailer_previews/case_mailer_preview.rb
```diff
@@ -0,0 +1,26 @@
+class CaseMailerPreview < ApplicationMailerPreview
+  def new_case
+    CaseMailer.new_case(get_case)
+  end
+
+  def comment
+    my_comment = CaseComment.first || FactoryBot.build(:case_comment)
```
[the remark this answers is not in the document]
jamesremuscat: The factory I wrote was broken :( I fixed this in another branch in 040934b59891d17e23d648255b7103f64124298d; I'll cherry-pick that commit into this one.

app/mailers/case_mailer.rb
jamesremuscat: The difference between here and `new_case` is in who gets excluded from the email that gets sent. I actually decided that a good place for all the "who gets this sort of case email" logic was in the mailer, rather than having it spread across comments, cases, maintenance window events, and whatever else we introduce in the future.

I can't see how to move it into `Case#email_recipients` in a way that would work everywhere we use it - if you have ideas, please let me know!

app/models/case_comment.rb
jamesremuscat: Have added a test in `spec/mailers/case_mailer_spec.rb`.

app/mailer_previews/case_mailer_preview.rb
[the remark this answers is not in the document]
bobwhitelock: Thanks :slightly_smiling_face:. Looks like there were a couple of other issues with how we were defining/using factories in general which made this still error when there was no saved CaseComments already, so I've pushed a couple of fixes for those.

app/mailers/case_mailer.rb
bobwhitelock: IIRC I was thinking that the logic here seems like 'the people who should be emailed about an update to a Case are all the contacts associated with the Case's Site, minus the user who initiated the action which is being emailed about', which could be captured in code, either in Case or CaseMailer, as something like (not actually tried this):

```ruby
def email_recipients_for(emailable_thing)
  site.all_contacts
    .map(&:email)
    .reject { |email| email == emailable_thing.user&.email }
end
```

(writing this out like this also makes me wonder if we should not email the user who initiated a MW transition, if any, as this would seem consistent with the other emails).

In any case possibly this is an over-abstraction right now, but maybe something to consider once we add another Case event which we want to email about, if this fits the same pattern.

bobwhitelock:
LGTM, thanks for bearing with my comments :shipit:

PR #166, opened 2018-04-17
Show maintenance on cases

jamesremuscat:
This PR adds maintenance window state transition events to the comments (now relabelled "case history") section on a case's `show` page.

These events are shown with (almost) the same text as is used in emails (changes are minor and mostly due to the medium being used).

Comments have a "speech bubble" icon and maintenance events a "wrench". The latter use the name of the user triggering the event where applicable, or `Flight Center` for automatic events. As discussed on Trello, we might want to iterate on how these are displayed in the future.

Trello: https://trello.com/c/rDJcEZNd/228-show-maintenance-in-case-feed


app/models/case.rb
```diff
@@ -122,6 +122,14 @@ def ticket_completed?
     COMPLETED_TICKET_STATUSES.include?(last_known_ticket_status)
   end
 
+  def events
+    (
+      case_comments.select(&:created_at) +  # We create an empty CaseComment to pass to our view, so there's one knocking
+      # around with no created_at as it's never saved
```
bobwhitelock: Sorry I don't fully follow this comment - could you expand on what you're referring to/point me in the direction of the empty CaseComment you mention?

Also a very pedantic aside: can we try to keep lines under 100 or ideally 80 characters for easier reading :slightly_smiling_face:

app/views/cases/_event.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="card event-card">
+  <div class="card-header d-inline-flex justify-content-between align-items-center">
+    <span><%= date.to_formatted_s(:long) %> - <b><%= name %></b></span>
+    <%= icon(defined?(type) ? type : 'info-circle') %>
```
bobwhitelock: Do we need a default for `type`? AFAICS we're not using this at the moment, and it would be useful to get an error rather than a default if we ever don't pass a `type` to this template, which will force us to decide what icon we want to show for that event.

spec/features/cases_spec.rb
```diff
@@ -145,3 +145,29 @@
...
+
+    visit case_path(open_case, as: admin)
+
+    event_cards = all('.event-card')
+    expect(event_cards.size).to eq(2)
+
+    expect(event_cards[0].find('.card-body').text).to eq('First')
+    expect(event_cards[1].find('.card-body').text).to eq('Second')
+
+  end
```
bobwhitelock: A test or two that cards for maintenance window transitions are shown when needed and display correctly would be useful, as currently nothing is exercising that code to ensure it works/keeps working.

app/views/maintenance_windows/transitions/_request.html.erb
```diff
@@ -0,0 +1,4 @@
+Maintenance requested for <%= transition.associated_model.name %> from
+<%= transition.requested_start %> until <%= transition.expected_end %> by <%= transition.window.requested_by.name %>;
+to proceed this maintenance must be confirmed on
+<%= link_to 'the cluster dashboard', transition.cluster_dashboard_url %>.
```
bobwhitelock: Is there a way we can prevent duplicating generating similar/the same text between these templates and the MaintenanceNotifier? I can see these either getting out-of-sync, or we have to always remember to update both places when we want to change this text.

AFAICS the only real difference is that in these templates we want to generate HTML links, while in MaintenanceNotifier we want some text along with a URL, so possibly either MaintenanceNotifier or a class extracted from this could generate the text, and provide a way to configure which of these should be generated, or alternatively the same templates could be shared between these with just the MWST and the text for any varying sections passed in.

jamesremuscat: Is this problem any different from having text and HTML versions of erb templates, and having to remember to update both versions every time you make a change? Possibly, because here the text versions are in `MaintenanceNotifier` and the HTML are in erb. Would using Rails' template rendering for both be a better situation?

Actually, if we had both HTML and text content for each type of transition, then the emails that get sent could just use those, meaning links get handled appropriately for the medium, so it probably is an improvement.

The "two places" thing still stands, though. I guess it would be possible to write a "take this fragment of HTML and do your best to make it plain text" function, but is it worth it? How often are we expecting these messages to change?

app/views/cases/_event.html.erb
bobwhitelock: Thanks for making the change - I tend to favour "fail fast (in development) so we're forced to decide how we want to handle the new case" but agree it's very situation-dependent :slightly_smiling_face:

spec/features/cases_spec.rb
bobwhitelock: Thanks for including this in the test - having code which is never touched by tests was just making me slightly worried that we could later change something seemingly completely unrelated and then have things break for users in production.

app/views/maintenance_windows/transitions/_request.html.erb
bobwhitelock: Good point that this is similar to what we (and most Rails apps) already do for text vs HTML emails. 

I have thought before in other apps when needing to keep two versions of the same email in sync that this is not ideal; I believe I have seen Gems out there before which will generate one of these email formats from the other, or both from some common source - not looked into these in any detail though so not sure how well they work.

I guess it does seem slightly different from that case to me as the email duplication is a standard thing most people working on a Rails app will be aware of, whereas you would need to be aware of this particular duplication specific to our app in order to remember to keep these in sync. Possibly moving the generation of both of these to ERB templates would be an improvement for this as you suggest, particularly given your point about then also being able to use the appropriate partial in emails.

I guess it also seems a bit less of a general problem in this case as here (at least at the moment) we are generating text with embedded `<a>` tags vs the same text with embedded URLs, whereas with emails the problem is generating arbitrary HTML vs nicely formatted text versions of the same content. So deduplicating these in our specific case seems like it could be simpler than solving the general problem in all cases.

In any case I don't think it's a priority right now, just something to be aware of; I've added a related card so we are aware of it if needed at some point in the future at https://trello.com/c/0hDF2EUP.

bobwhitelock:
All LGTM now :shipit:

PR #176, opened 2018-04-24
Track case state internally

jamesremuscat:
This PR adds our own concept of a case's state to Flight Center, meaning we no longer rely on RT's ticket statuses.

When first created, a `Case` has a state of `open`. It remains open until marked as `resolved` by an Alces admin, at which point new comments cannot be added and the only subsequent state is `archived`, signifying that the case has been accounted for in the credit reporting process (i.e. Claire's spreadsheet). Resolved cases cannot be reopened, and archived cases cannot be unarchived.

Note that this is a different meaning of the word 'archived' to that previously used by Flight Center. It had previously been used solely as a flag for toggling display of cases in the main list on or off. Now, only open cases are shown in the main list, and all cases are shown in the "archive" list.

State changes are logged, using the same mechanism as for maintenance windows, and displayed in a case's history feed on the case page.

Much of the existing RT framework needs to remain until at least after the data migration has been run (since we'll need to pull RT ticket status one last time). I've tried to mark everywhere affected with `@deprecated` (not including the obvious RT interface classes) and we should delete these once we've released this set of changes. I've also left the "RT Ticket" field displayed on a case's page since this might be useful during the transition.

Trello: https://trello.com/c/P4QgU9M4/232-add-status-tracking-directly-to-cases


bobwhitelock:
FYI I've rebased this on `develop` to remove the duplicated commits and merges. `git diff` tells me the changes remain identical, but the history is now much simpler.

spec/factories.rb
```diff
@@ -1,5 +1,13 @@
 
 FactoryBot.define do
+  factory :case_state_transition do
+    add_attribute(:case) { create(:open_case) }
```
bobwhitelock: There can be issues with defining relations like this which I only discovered recently, as this always creates the Case whether we want to or not. We may want to use `association` instead, see https://github.com/alces-software/alces-flight-center/commit/ba0df6ad0ad6b5f2d20ff3e98e8d5e7002158bc6.

spec/features/log_spec.rb
```diff
@@ -65,11 +65,34 @@ def select_details_for(input_cases)
...
+
+      # Revisit the page so given Case would be shown, if we do not
+      # successfully filter it out, to avoid false positive.
+      visit current_path
+
+      expect do
+        select resolved_case.subject
+      end.to raise_error(Capybara::ElementNotFound)
+    end
+
```
bobwhitelock: Good that we're testing this case too - looks like we could able to avoid duplicating this from the test below by doing something like this:

```ruby
[:resolved, :archived].each do |state|
  # the test as they both are now, with just the parts dependent on state
  # changed to use the `state` vaiable
end
```

spec/features/maintenance_windows_spec.rb
```diff
@@ -231,12 +231,29 @@ def duration_input_group
         expect(invalid_feedback).to have_text('Must be greater than 0')
       end
 
+      it 'cannot select resolved Case for Cluster to associate' do
```
bobwhitelock: As above, can avoid duplicating this by iterating through the statuses we want to test and defining a slightly more generic test once.

bobwhitelock:
@jamesremuscat Actually the next changes I'll make may touch the same parts of the code. Since you've also got more changes based on this I'm going to merge this through now, to avoid us having several diverging versions of things. I've just added a card at https://trello.com/c/vLvITAXP/248-look-at-comments-on-pr-176 to remind us to come back and look at these comments in a future PR.

app/decorators/case_state_transition_decorator.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  private
+
+  def properties_for_event(event)
+    case event
+    when 'resolve'
+      return 'This case was marked as resolved.', 'check-circle-o'
+    when 'archive'
+      return 'This case was archived.', 'archive'
```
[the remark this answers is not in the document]
jamesremuscat: We need explicit `return`s since we're returning a tuple (syntax error without it).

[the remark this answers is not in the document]
bobwhitelock: Ah TIL, I normally use an array in this case but I guess I should expect Ruby to have several ways to do everything :slightly_smiling_face:

PR #179, opened 2018-04-26
Disable non-consultancy Case comments

bobwhitelock:
This PR disables commenting on Cases for Site contacts when the Case is not consultancy (with our business reasoning for doing this being that theoretically the original Case should contain all relevant information in this situation). 

@jamesremuscat would you be able to take a quick look at this, since this touches the stuff you've been working on? :slightly_smiling_face: 

Trello: https://trello.com/c/Wqcy44Py/233-disable-case-commenting-with-standard-reason-displayed-for-tier-1-2-cases-1-4-day

### DIFF
--- a/app/decorators/case_decorator.rb
+++ b/app/decorators/case_decorator.rb
@@ -69,4 +69,8 @@ class CaseDecorator < ApplicationDecorator
     description = TIER_DESCRIPTIONS[tier_level]
     "#{tier_level} (#{description})"
   end
+
+  def commenting_disabled?
+    current_user.contact? && !consultancy?
+  end
 end
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -184,6 +184,10 @@ class Case < ApplicationRecord
     Utils.rt_format(case_properties)
   end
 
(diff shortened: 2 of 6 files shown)


jamesremuscat:
All looks good, but I wonder if we could make the "this is a non-consultancy issue" message more obvious - perhaps as an additional "event" card added above the comment box? It's not immediately obvious why you can't type in it, and I wouldn't instinctively mouse-over the submit button to work out why.

(In isolation, it's also not obvious that the textarea _is_ disabled - until you've seen a non-disabled one to compare colours with.)

I understand that wanting to still display the disabled textarea is a business requirement ;)

bobwhitelock:
Thanks for reviewing, and I think that suggestion sounds like it would be a good improvement. In the interest of getting things ready to be released to staging/prioritising what makes it into the next release I'm not going to address it immediately, but I've added a card for it at https://trello.com/c/oSqqaXvS/250-make-it-more-obvious-why-non-consultancy-cases-cannot-be-commented-on and will come back and look at it soon.

PR #180, opened 2018-04-26
Assign users to cases

jamesremuscat:
This PR adds the concept of an "assignee" to `Case`s, and provides UI for admins to set or clear an assignee for a case.

The current assignee of a case is displayed in both the cases table and the case show page. Only admins are given UI for changing the assignee of a case. A case may be assigned to any contact that is a member of the same site as the case, or to any admin, or to nobody.

Change of assignee is recorded by the `audited` gem, and displayed in the case history feed alongside other events. When an assignee is changed, an email is sent to all contacts of the appropriate site (plus the admin to whom the case is being assigned, where appropriate).

This work was originally done in `feature/assign-cases-to-users` but trying to rebase or merge that with `develop` was a nightmare, so I cherry-picked into this fresh branch instead.

Trello: https://trello.com/c/6rxwL4sG/244-add-user-assignment-to-cases

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -72,8 +72,13 @@ class CasesController < ApplicationController
   end
 
   def assign
-    new_assignee = User.find(params[:assignee_id])
-    change_action "Support case %s assigned to #{new_assignee.name}." do |kase|
+    new_assignee_id = params[:case][:assignee_id]
+    new_assignee = new_assignee_id.empty? ? nil : User.find(new_assignee_id)
+    success_flash = new_assignee ?
+                        "Support case %s assigned to #{new_assignee.name}."
+                        : 'Support case %s unassigned.'
+
+    change_action success_flash, redirect_path: case_path do |kase|
       kase.assignee = new_assignee
     end
   end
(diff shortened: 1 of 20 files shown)


app/controllers/cases_controller.rb
```diff
@@ -91,7 +103,7 @@ def case_params
     )
   end
 
-  def change_action(success_flash, &block)
+  def change_action(success_flash, redirect_path: case_path, &block)
```
bobwhitelock: Possibly I'm missing something - we now have `redirect_path` as a parameter, but the only time it looks like it's passed (in `assign`) it is given the same value as the default. So is it needed?

app/decorators/audited/audit_decorator.rb
```diff
@@ -0,0 +1,47 @@
...
+    private
+
+    def change_card(date, user, change)
+      field = change[0]
+      from, to = *change[1]
+
+      type = send("#{field}_type")
+      text = send("#{field}_text", from, to)
+
+      render_card(date, user&.name || 'Flight Center', type, text)
```
bobwhitelock: Do we need the `user&.name || 'Flight Center'`? Can the user ever be nil when we get here?

PR #220, opened 2018-05-02
Remove last unnecessary ticket and ticket_id references

bobwhitelock:
As just discussed IRL these are the changes to be more consistent and correct with how we refer to cases. Essentially this makes it so:

- we always refer to Cases when we mean Flight Center Cases, rather than tickets, and only refer to tickets when we actually mean an RT ticket;

- we use the `display_id` method wherever we previously showed the RT ticket ID (except where we need to keep our previous behaviour for a reason, such as in emails to avoid breaking threading with previous emails);

- `display_id` has been moved to `Case` since it fits better there now;

- the comments in https://github.com/alces-software/alces-flight-center/commit/5402a790f415a474ceb63ca568b0fd539b1240ab may be relevant to you @jamesremuscat for your current work on https://trello.com/c/dzY3fb5C/252-new-case-ids-ready-for-use-after-replacing-rt, as we may want to update these places then (also see https://github.com/alces-software/alces-flight-center/blob/5402a790f415a474ceb63ca568b0fd539b1240ab/app/models/case.rb#L205-L206, which may need updating or become obsolete depending on how we implement things).

### DIFF
--- a/spec/decorators/application_decorator_spec.rb
+++ b/spec/decorators/application_decorator_spec.rb
@@ -54,41 +54,41 @@ RSpec.describe ApplicationDecorator do
 
       it 'includes correct icon when has requested maintenance window' do
         window = create(:requested_maintenance_window, component: component)
-        ticket_id = window.case.rt_ticket_id
+        display_id = window.case.display_id
 
         expect(subject).to include(
           h.icon(
             'wrench',
             inline: true,
             class: 'faded-icon',
-            title: "Maintenance has been requested for #{component.name} for ticket #{ticket_id}"
+            title: "Maintenance has been requested for #{component.name} for case #{display_id}"
           )
         )
       end
 
       it 'includes correct icon when has confirmed maintenance window' do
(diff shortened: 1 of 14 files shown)


spec/decorators/application_decorator_spec.rb
```diff
@@ -54,41 +54,41 @@
 
       it 'includes correct icon when has requested maintenance window' do
         window = create(:requested_maintenance_window, component: component)
-        ticket_id = window.case.rt_ticket_id
+        display_id = window.case.decorate.display_id
```
jamesremuscat: Do you still need `decorate` here (and in subsequent tests)?

bobwhitelock: No good catch, I'll rebase with this fixed - was leftover from before I decided `display_id` fit better in the model directly :slightly_smiling_face:

bobwhitelock:
Thanks for looking over, will merge once CI passes

PR #221, opened 2018-05-02
Generate per-cluster sequential case IDs

jamesremuscat:
This PR changes the way we publicly refer to `Case`s by introducing a per-cluster sequential "display ID", prefixed with a new "shortcode" allocated to each cluster.

The `#display_id` method on `Case` is therefore replaced by simply accessing the value of this stored field.

New-style display IDs are of the form `BAR123` - this being the 123rd case raised for the `BAR` cluster (which, in production, would be the shortcode allocated to Barkla).

Old cases with RT ticket IDs use a display ID of the form `RT12345`, as previously.

Display IDs are allocated immediately before the initial save of a `Case` (specifically, using the `before_create` callback). It is from `Cluster` that the numeric part of the display ID is obtained, and we use a database row-level lock on the cluster to ensure concurrently created cases comply correctly with the uniqueness constraint. 

Since the new display IDs include a cluster hint, we no longer include the cluster name explicitly in email subject lines.

We also support accessing cases via their display IDs in the URL. This works case-insensitively so that `/cases/bar1` and `/cases/BAR1` both go to the same place. Accessing cases by their database ID still works, so old links will continue to function, but we no longer display those links anywhere. Code should use `Case#find_from_id` rather than `Case.find` to take advantage of this omni-findyness.

There are two data migrations included in this PR:

- One to allocate a shortcode to all clusters. This uses the first three characters in the cluster name, upper-cased; except for "Demo Cluster" which uses `DEMO`. There's some attempt at deduplication which almost certainly won't be needed, except possibly in development if you have old data lying around (I did).

- One to assign display IDs to cases. For cases with an RT ticket ID, a display ID of the form `RTxxxx` is created. For cases without (and there won't be any in live) the cluster's `case_index` is used for numbering.

I've also taken the liberty of making the ID column in the cases table a clickable link as per https://trello.com/c/yv5pDErI/261-cases-ticket-id-needs-to-be-on-the-left-and-make-it-a-link.

Trello: https://trello.com/c/dzY3fb5C/252-new-case-ids-ready-for-use-after-replacing-rt

### DIFF
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -172,8 +172,7 @@ class Case < ApplicationRecord
     if rt_ticket_id
       "[helpdesk.alces-software.com ##{rt_ticket_id}]"
     else
-      # TODO Update this with identifier in format `BAR123` once this exists.
-      "[Alces Flight Center ##{id}]"
+      "[Alces Flight Center #{display_id}]"
     end
   end
 
@@ -185,9 +184,16 @@ class Case < ApplicationRecord
     # for an explanation). If we want to change this format and avoid this
     # consequence then a solution would be to first add a new field for this
     # whole string, and save and use the existing format for existing Cases.
-    # TODO should we update this to include identifier in format `BAR123` once
-    # this exists?
-    "#{cluster.name}: #{subject} [#{token}]"
+    #
+    # FSR using the conditional in `#email_identifier` rather than repeating it
+    # here causes Rails to Base64-encode the plain text part of the email, which
+    # causes some of our tests to fail...
+    if rt_ticket_id
+      "#{cluster.name}: #{subject} [#{token}]"
+    else
+      # With 'new' display IDs we have a cluster hint, so don't include it twice.
+      "#{subject} [#{token}]"
+    end
   end
 
   def email_properties
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -86,7 +86,7 @@ class Case < ApplicationRecord
 
   before_validation :assign_default_subject_if_unset
 
-  after_create :set_display_id
+  before_create :set_display_id
   after_create :send_new_case_email
   after_update :maybe_send_new_assignee_email
 
@@ -297,20 +297,19 @@ class Case < ApplicationRecord
 
   def set_display_id
     return if self.display_id
-    # Note: this method is called AFTER create, to ensure that the case is valid
-    # before we increment the cluster's `case_index` field. Otherwise display IDs
-    # could end up non-sequential.
+    # Note: this method is called `before_create`, which is AFTER validation is run.
+    # This ensures that the case is valid before we increment the cluster's
+    # `case_index` field. Otherwise display IDs could end up non-sequential.
 
     if self.rt_ticket_id
       self.display_id = "RT#{rt_ticket_id}"
     else
       self.display_id = "#{cluster.shortcode}#{cluster.next_case_index}"
     end
-    save!
   end
 
   def has_display_id_when_saved
-    # We want to be able to save the case initially without a display id
+    # We want to be able to validate the case initially without a display id
     errors.add(:display_id, 'must be present') unless !persisted? or display_id
   end
 
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -101,11 +101,11 @@ class Case < ApplicationRecord
     self.display_id.parameterize.upcase
   end
 
-  def self.find_from_id(id)
+  def self.find_from_id!(id)
     if /^[0-9]+$/.match(id)  # It's just a numeric ID
       Case.find(id).decorate
     else # It has non-digits in - let's assume it's a display ID
-      Case.find_by_display_id(id.upcase)
+      Case.find_by_display_id!(id&.upcase)
     end
   end
 
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -204,12 +204,6 @@ class Case < ApplicationRecord
     super(new_assignee)
   end
 
-  def display_id
-    # TODO Once https://trello.com/c/dzY3fb5C has been implemented we should
-    # replace `##{object.id}` with that identifier for non-RT tickets.
-    rt_ticket_id ? "RT#{rt_ticket_id}" : "##{id}"
-  end
-
   private
 
   # Picked up by state_machines-audit_trail due to `context` setting above, and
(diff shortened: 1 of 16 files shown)


bobwhitelock:
When I pull down this branch and run `rake alces:data:import_and_migrate_production` (which is the closest we can get to what the effect would be if the code on this branch was deployed right now to production) I get errors from a data migration which has not yet run in production:

```
== 20180409163843 MigrateExistingDataToAccountForTiers: migrating =============
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

Validation failed: Display has already been taken, Display must be present
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:129:in `block in add_level_1_tier_fields_to_support_type_change_cases'                                       
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:171:in `block (2 levels) in each_case'                                                                       
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:170:in `block in each_case'                                                                                  
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:169:in `each'                                                                                                
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:169:in `each_case'                                                                                           
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:127:in `add_level_1_tier_fields_to_support_type_change_cases'                                                
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:17:in `update_cases'                                                                                         
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:4:in `up'                                                                                                    

Caused by:
ActiveRecord::RecordInvalid: Validation failed: Display has already been taken, Display must be present
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:129:in `block in add_level_1_tier_fields_to_support_type_change_cases'                                       
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:171:in `block (2 levels) in each_case'                                                                       
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:170:in `block in each_case'                                                                                  
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:169:in `each'                                                                                                
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:169:in `each_case'                                                                                           
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:127:in `add_level_1_tier_fields_to_support_type_change_cases'                                                
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:17:in `update_cases'                                                                                         
/home/bob/src/alces-software/alces-flight-center/db/data_migrations/20180409163843_migrate_existing_data_to_account_for_tiers.rb:4:in `up'                                                                                                    
Tasks: TOP => alces:data:import_and_migrate_production => data:migrate
(See full trace by running task with --trace)
```

My initial thought is that it looks like we're validating the uniqueness and presence of `display_id`, but none of the cases at the time we're running the preceding migrations have this set. Possibly we don't want to validate the presence of `display_id` yet, and just the uniqueness only if it is set (as with `rt_ticket_id`). We can make a note to later come back and tighten up this validation once we've deployed this, as with some other similar changes we'll need to make then.

As a more general suggestion I'd suggest running this task every so often, particularly after adding new migrations or data migrations, which I think has the benefits of both ensuring you're working with data as close to production as possible and ensuring the un-applied migrations will successfully run when run as they will be in production (with all the Rails migrations followed by all the data migrations). In turn this means when writing data migrations you can both better account for the real data as it currently is and safely avoid having to handle situations which theoretically could exist but actually don't (which it looks like you have done to some extent in your data migrations here).

jamesremuscat:
In case you didn't spot this @bobwhitelock - I've taken your suggested approach of making the validation more lax for the time being.

I've tagged those sections with `XXX` as I see there's a pre-release task to search for that, to make sure it doesn't get forgotten

app/models/case.rb
```diff
@@ -84,11 +91,24 @@ class Case < ApplicationRecord
...
 
+  def to_param
+    self.display_id.parameterize.upcase
+  end
+
+  def self.find_from_id(id)
+    if /^[0-9]+$/.match(id)  # It's just a numeric ID
+      Case.find(id).decorate
+    else # It has non-digits in - let's assume it's a display ID
+      Case.find_by_display_id(id.upcase)
```
bobwhitelock: I think we should use the `find_by_display_id!` method here, as this will mean we fail fast with the usual `RecordNotFound` error if we attempt to find a non-existent Case (`Case.find` does this by default I believe so that branch is fine).

I think using the `find_by_*!` version of methods is usually what we want (unless we want to explicitly check for a `nil` record at the point we use the finder method), as that means we can either guarantee we will get an error raised or an AR object; using the non-`!` version means we can get `nil` which can then cause a `NoMethodError` in a distant place. In particular in this case this means I get
```
NoMethodError in CasesController#show
undefined method `associated_model' for nil:NilClass
```
when I navigate to a non-existent Case like `http://localhost:3000/cases/bar123`

If we do make this change I think it would also make sense to rename this method to `find_from_id!`, to match the AR convention since it is a thin wrapper around calling the correct AR method.

jamesremuscat: Good shout - both these changes work in concert together, so we no longer need the conditional in `assign_scope`.

bobwhitelock:
@jamesremuscat thanks for making that change - is this ready to be merged now?

jamesremuscat:
:shipit:

PR #225, opened 2018-05-04
Support help text for Tier fields

### LINKED ISSUE #211: Add "help text"/"description" for tier 2 fields
It would be clearer to users how to fill out tier 2 forms if there was a bit of (optional) descriptive text that could be associated with each field that explains what kinds of values are expected.

This will probably also be required for tier 1 forms, but may be built into the app rather than via the database (TBD).

bobwhitelock:
This PR allows specifying `help` text within the Tier fields JSON, and having this then be displayed in the Case form alongside the field. This is done by refactoring and then adding this functionality, in such a way that it should be slightly easier to add similar additional fields with special handling to the Tier fields JSON next time.

This closes #211 and completes https://trello.com/c/2RkPP0zI/270-add-ability-to-add-hints-for-inputs-in-tiers-in-case-form.

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -98,7 +98,7 @@ class CasesController < ApplicationController
       :service_id,
       :subject,
       :tier_level,
-      fields: [:type, :name, :value, :optional],
+      fields: [:type, :name, :value, :optional, :help],
     )
   end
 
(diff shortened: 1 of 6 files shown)


bobwhitelock:
Also: I've added some example help text for some of the fields of a single Issue/Tier, the level 2 Tier for the 'From available Alces Gridware' Issue; someone with knowledge of what help text we want for other fields should come back and add/amend this before we release this (see https://trello.com/c/NRiaksur/274-go-through-and-add-remove-help-text-for-tier-fields-as-needed).

PR #226, opened 2018-05-04
Add design doc for implementing Tier 1 tools

bobwhitelock:
I wrote some words on how I plan to implement the MOTD update and user/group management Tier 1 tools, would appreciate any comments/questions/suggestions/pointing out of possible flaws any of you might have :slightly_smiling_face:. Related card: https://trello.com/c/TKgqhL7x/271-request-update-to-motd-tier-1-tool.

### DIFF
--- a/docs/design/01-initial-tier-1-tool-planning.md
+++ b/docs/design/01-initial-tier-1-tool-planning.md
@@ -1,0 +1,133 @@
+
+Plans for initial MOTD tool, with consideration to future user/group management
+tool(s).
+
+# Requirements
+
+- Every Cluster should have a MOTD.
+
+- We want to update this via the Case form; unlike with any existing Tier
+  fields when this is done we also want to show the current MOTD in the form.
+
+- It is possible that we will get multiple MOTD change request Cases in
+  succession, in which case only the last change really needs to be made.
+
+- Soon we will want similar handling for creating, updating, and deleting
+  Cluster users for a Cluster via the Case form.
+
+
+# Implementation plan
+
+- Add `motd` field to Cluster.
+
+- Add `tool` field to Tier, which will be required for level 1 Tier/forbidden
+  for other levels.
+
+- Add a new 'Request change of MOTD' Issue with a level 1 Tier with a new
+  `motd` tool.
+  - This might fit best within our current hierarchy to be available under the
+    'HPC Environment' Service in the 'End User Assistance' Category (or
+    wherever else you want).
+
+- When a level 1 Tier is selected in the Case form the associated tool will be
+  displayed
+  - for the MOTD tool this will just mean:
+    - the current MOTD is displayed;
+    - a textarea is given pre-filled with the current MOTD, which can then be
+      adapted/replaced with the new one to be requested.
+
+- When a Case is created using the tool, in addition to the usual things done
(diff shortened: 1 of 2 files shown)


ste78:
just a quick clarification question... Will the motd change tool also show the pending changes? ie all the open motdchange requests?

mjtko:
I'm a bit confused about the term `apply` -- I'm visualizing an "Apply" button for the Admin to click on, which makes me think it's going to "do" something rather than the Admin having already done it themselves.

Perhaps this terminology may be better as `action`/`actioned`? The Admin clicks on an "Actioned" button?

(I'm also envisaging a potential future where we may have an "In progress" -> "Actioned" -> "Resolved" workflow).

mjtko:
(Or, thinking aloud regarding the idea of opening multiple conflicting/overriding requests possibly "Open" -> "Superseded".  But later. :))

bobwhitelock:
@ste78: 
> just a quick clarification question... Will the motd change tool also show the pending changes? ie all the open motdchange requests?

I wasn't planning to do this initially as I thought the situation where a user creates a Case to request a change to the MOTD while they already have one open, and to not be aware of this because they just created it, is not likely to be common enough to be worth adding extra handling for this (the main use case I can think of for this is if they create the Case and then realise they want to change it further/back). But if you'd like this I can do this.

@mjtko:
> I'm a bit confused about the term `apply` -- I'm visualizing an "Apply" button for the Admin to click on, which makes me think it's going to "do" something rather than the Admin having already done it themselves.
>
> Perhaps this terminology may be better as `action`/`actioned`? The Admin clicks on an "Actioned" button?

I thought of using this terminology as within the context of Flight Center it will "do" something, as clicking `apply` will make the necessary changes within Flight Center's database - i.e. in this case update the `motd` for the Cluster, and in future cases create ClusterUsers etc. I didn't think this would cause confusion as currently there are no plans to integrate Flight Center so it actually directly makes changes to physical Clusters themselves (e.g. actually creating user accounts directly), so I didn't think this would make admins think they don't actually need to log into the Cluster and make the change. I can call this `action` if you prefer though - to me `action` possibly sounds like it won't update data, and possibly implies that it isn't repeatable, but either way we can make clear what `apply`/`action` means to admins by explaining this/making it clear in the UI.

bobwhitelock:
@mjtko - following on from the above/what we talked about yesterday, what do you think about the following:

1. I use whatever seems to fit best for `action`/`apply`/etc. internally;

2. since there doesn't seem to be a single word which perfectly captures what this button will do, we make this explicit in the button text, something like:
  - 'before the button has ever been clicked' = green button with text "Apply change - I have made this change in the cluster";
  - 'after it has been successfully clicked' = orange button with text "Already applied - attempt to re-apply change"?

mjtko:
Internally it can use whatever makes sense from the code perspective.

From an Admin/UI perspective, I'd be expecting to see buttons that represent the actions I have taken or am in the process of taking, perhaps something like this:

"Done (change has been applied to system)"
"Redo (change has been reapplied to system)"

We probably don't need to decide exactly what it says right now however, and I'm happy to go with whatever you feel fits best.

bobwhitelock:
Going to merge this since I've started implementing this, and this just serves as a record of the initial design. Any further comments/suggestions/requests are still welcome though here or via the usual routes.

PR #227, opened 2018-05-08
Feature/track time on cases

jamesremuscat:
This PR adds the ability to track total time worked against a case.

Visible only to admins, each case now displays a "time worked" value in hours and minutes. For open cases, this value can be set (absolutely, not cumulatively - it should always be set to the *total* time worked to date).

Internally, it is stored as integer minutes, converted to a more friendly hours/minutes for display and entry.

Changes to `:time_worked` are stored by the `audited` gem in the same way as assignee changes, and events displayed in a case's history (again, only for admins).

To enforce process, the value of `:time_worked` may not be changed when a case is `resolved` or `closed`.

Trello: https://trello.com/c/aCQmt0eg/266-bonus-add-time-tracking-to-cases-for-admins

### DIFF
--- a/app/views/cases/_time_worked.html.erb
+++ b/app/views/cases/_time_worked.html.erb
@@ -1,0 +1,23 @@
+<% if current_user.admin?
+  hours = @case.time_worked % 60
+  minutes = @case.time_worked - (60 * hours)
+%>
+<tr>
+  <th>Time worked</th>
+  <td colspan="3">
+    <%= form_tag({ controller: 'cases', action: 'set_time' }, method: :post, class: 'form-inline') do %>
+      <div class="form-group">
+        <label>
+          <%= number_field :time, :hours, class: 'form-control mr-1', min: 0, value: hours %>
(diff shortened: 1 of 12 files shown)


bobwhitelock:
Rebased on `develop` for to resolve conflict and for easier reviewing.

app/views/cases/_time_worked.html.erb
```diff
@@ -0,0 +1,46 @@
...
+                           max: 59,
+                           value: minutes
+          %>
+          minutes
+        </label>
+      </div>
+      <% if @case.time_entry_allowed? %>
+        <%= submit_tag 'Change time worked',
+                       class: 'form-control btn-primary ml-2',
+                       disabled: !@case.time_entry_allowed?
```
bobwhitelock: Unless I'm mistaken you don't need this attribute, which will never be true as the `submit_tag` is only rendered when `@case.time_entry_allowed?` is true.

PR #229, opened 2018-05-08
Display logs tab on the component dashboard

atoghill:
Much like on the `Cluster` dashboard there is now a `Logs` tab on the `Component` dashboard. This allows you to view all logs for that component from within its dashboard.

This is based off #228


bobwhitelock:

bobwhitelock:
Looks great, thanks. Only minor thing here is can you link back to the associated Trello card just so it can be found when viewing the PR, otherwise this is great :slightly_smiling_face:.

PR #230, opened 2018-05-09
Create a case from the drop-down menu

atoghill:
A new case can now be created from the `Cases` drop-down menu using the `Create` option. This functions exactly the same as the old button on the cases page but is quicker to access if you aren't already on the cases page. 

The `case_form_button` has been completely removed for now but it may be a good idea to have a nicer button somewhere on the cases page as another way to create a new case. There is a chance users would not know to look under the drop-down for this option if they only go to create a case from the cases page directly. A simple create button makes the process potentially more intuitive but for now removing it may be fine.

See https://trello.com/c/PEgWH3vT/276-move-case-creation-to-entry-in-tab-menu for more info


bobwhitelock:
Looks great, thanks

PR #231, opened 2018-05-09
Display the case subject on the case link

atoghill:
The case subject is now displayed on the tooltip of any case link.

_(Trello Card: https://trello.com/c/lcqWAgVt/277-show-ticket-title-in-some-way-in-maintenance-table)_


bobwhitelock:
Looks great, thanks - I've squashed and merged the two commits in the above commit, so that we don't have the failing test in between them, so I'll close this PR since it's been merged just not picked up by GitHub.

Not a major deal, but basically the less commits we have where tests are failing the more useful tools like `git bisect` can be, by allowing us to go back in time to different points when debugging and still have things otherwise in a working state. For future reference in a similar situation to this in future you could use `git commit --amend` when creating the second commit so the new changes are added to the previous commit, rather than creating a new one.

PR #236, opened 2018-05-09
Maintenance tab within Component and Service dashboards

atoghill:
This PR adds the `Maintenance` tab to the `Component` and `Service` dashboards, functioning just like the same tab within the `Cluster` dashboard.

_(Trello Card: https://trello.com/c/LuMoeHAR/281-show-maintenance-table-on-component-and-service-dashboards)_

### DIFF
--- a/app/models/component.rb
+++ b/app/models/component.rb
@@ -20,6 +20,15 @@ class Component < ApplicationRecord
 
   after_create :create_component_expansions_from_defaults
 
+  def unfinished_related_maintenance_windows
+    component = [self]
+    component
+      .map(&:maintenance_windows)
+      .flat_map(&:unfinished)
+      .sort_by(&:created_at)
+      .reverse
+  end
+
   private
 
   def create_component_expansions_from_defaults
--- a/app/models/component.rb
+++ b/app/models/component.rb
@@ -20,15 +20,6 @@ class Component < ApplicationRecord
 
   after_create :create_component_expansions_from_defaults
 
-  def unfinished_related_maintenance_windows
-    component = [self]
-    component
-      .map(&:maintenance_windows)
-      .flat_map(&:unfinished)
-      .sort_by(&:created_at)
-      .reverse
-  end
-
   private
 
   def create_component_expansions_from_defaults
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -53,6 +53,10 @@ Rails.application.routes.draw do
     end
   end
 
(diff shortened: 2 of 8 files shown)


atoghill:
Rebased off of `develop` to fix the issue with Travis, see #238

app/models/component.rb
```diff
@@ -20,6 +20,15 @@ class Component < ApplicationRecord
 
   after_create :create_component_expansions_from_defaults
 
+  def unfinished_related_maintenance_windows
+    component = [self]
+    component
+      .map(&:maintenance_windows)
+      .flat_map(&:unfinished)
+      .sort_by(&:created_at)
+      .reverse
```
bobwhitelock: I can see why you've done this, as you're following a similar process to what we do in `Cluster#unfinished_related_maintenance_windows`, however in this case I think we can simplify this a bit.

In Cluster things are a bit more complex, as for a Cluster we want to show all the unfinished maintenance for both the Cluster and all the Components and Services which it is made of. In this case though we just want to show the maintenance for this Component, and
```ruby
    component = [self]
    component
      .map(&:maintenance_windows)
      .flat_map(&:unfinished)
```
is just a complex way of calling `component.maintenance_windows.unfinished` (since we're basically mapping twice over an array with one element). So I think we can simplify this to something like this:
```ruby
  def unfinished_related_maintenance_windows
    maintenance_windows.unfinished
      .sort_by(&:created_at)
      .reverse
  end
```

The same applies to the code for Service too, so rather than duplicating this we could also extract this to the shared `ClusterPart` concern which contains the code shared between these.

config/routes.rb
```diff
@@ -155,6 +156,7 @@
 
     resources :services, only: :show do
       resolved_cases.call
+      resources :maintenance_windows, path: :maintenance, only: :index
```
bobwhitelock: You could extract a Proc for the different places we do this, as we do with various other duplicated route definitions.

app/models/component.rb
atoghill: Yep this was something I was unsure of how to handle within this project and knew you would pick up on. Wasn't sure where shared code would go for this sort of thing but I'll move it out to that concern, cheers

config/routes.rb
atoghill: Agreed, that makes sense. Didn't even think about that but I have moved it to a proc now.

bobwhitelock:
Looks great, thanks

PR #237, opened 2018-05-09
Fix various leftover `archived` case state references

jamesremuscat:
As title.

We might want to merge this branch into the next release as well as `develop` - though I'd judge this to be a low-impact bug since it only affects closed cases (of which we have none yet) and then only display issues.


bobwhitelock:
Rebased on `develop` so includes the fix for the Travis issue (#238).

app/views/cases/index.html.erb
```diff
@@ -28,10 +28,10 @@ end %>
         # Hence logic simplifies to the below (!=)
       %>
           <% if @show_resolved != c.open?  %>
-            <% if c.hidden? && !manage_cases_page %>
+            <% if (c.resolved? || c.closed?) && !manage_cases_page %>
```
bobwhitelock: Can this condition ever be true now? Since if we're not on the `manage_cases_page` I believe we only show `open` Cases, and if we are on the `manage_cases_page` we only show non-`open` Cases. If so we could remove the whole conditional, and we could probably remove the `elsif` as well since it's just to do with the legacy charging logic. 

If this is correct, possibly you may want to do this in your current branches rather than here, since this is related to what you're working on at the moment and I don't think this is a new issue.

bobwhitelock: Ah I see - as a contact if I visit `/cases/resolved` this will be triggered. However we probably don't want this behaviour any more, since as we now only show `resolved` and `closed` Cases on this page this means every Case appears faded out, which is not that helpful. But since this is not a new issue/would conflict with many other branches currently being worked on/isn't a big deal we can address this later. Thanks for clarifying.

PR #238, opened 2018-05-09
Add flag to opt-in to using apt-get install

atoghill:
As of today we need to opt-in to using apt-get install as Travis has changed the build script to not run apt-get update by default. This may change in the future but this is their fix for the time being.


bobwhitelock:
Thanks for looking into the issue @atoghill, LGTM can be merged if/when CI passes

PR #241, opened 2018-05-10
Add credit charge to close cases

jamesremuscat:
This PR makes Flight Center require the entry of a "credit charge" value in order to move Cases from `resolved` to `closed`.

By default a Case will have a `credit_charge` of `nil` while in states `open` or `resolved`. At the point of `close`-ing a Case, the admin is now prompted for a positive integer value for `credit_charge` - this can be zero.

Credit charge, where present, is shown in the resolved Cases table and on the Case page.

I've also deleted the previous incarnation of "credit charge" features - this all happened in a single commit, e34a87f, in case you'd prefer to keep these around (though note the name of the previous association `credit_charge` on Case will now conflict with the field of the same name).

Trello: https://trello.com/c/q9CBLhgs/267-implement-charging-for-cases


bobwhitelock:
Rebased on `develop` to resolve conflicts and for easier reviewing.

jamesremuscat:
Fixed the test, and added an event card for adding a charge since it wasn't much work to do.

bobwhitelock:
Looks great, thanks @jamesremuscat :shipit:

PR #243, opened 2018-05-10
Fix various leftover `archived` case state references (backported for next release)

bobwhitelock:
Backporting PR #237 to go into the next release (rather than needing to wait until the one after that for this fix to be deployed).


bobwhitelock:
Also needed to cherry-pick in the fix from https://github.com/alces-software/alces-flight-center/pull/238 to get CI passing.

PR #245, opened 2018-05-10
Case escalation

jamesremuscat:
This PR adds the ability for users - both contacts and admins - to escalate an open case up to tier 3.

When viewing an open tier 2 case, there is now a button next to the "Tier" value marked "Escalate". Clicking this button will open a modal almost identical to the one used in the new-case form warning that: 

> Escalating a support case means that the case becomes potentially chargeable, and may incur a charge of support credits.

If the user proceeds, then the case is set to tier 3 and an entry logged in the case history feed.

This is a one-time, irreversible action - once done a case cannot be de-escalated (though whether or not a case actually incurs a credit charge is a separate, if related, concern). It is not possible to escalate resolved or closed tickets.

Trello: https://trello.com/c/vrSKJGx0/268-add-escalation-of-cases

### DIFF
--- a/app/decorators/audited/audit_decorator.rb
+++ b/app/decorators/audited/audit_decorator.rb
@@ -24,7 +24,9 @@ module Audited
       text = send("#{field}_text", from, to)
       admin_only = ADMIN_ONLY_CARDS.include?(field)
 
-      render_card(date, user&.name || 'Flight Center', type, text, admin_only)
+      if text
+        render_card(date, user&.name || 'Flight Center', type, text, admin_only)
+      end
     end
 
     def render_card(date, name, type, text, admin_only)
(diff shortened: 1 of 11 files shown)


bobwhitelock:
There's some conflicts between this and your other recently merged branches, would you be able to rebase to resolve? Thanks :slightly_smiling_face:

jamesremuscat:
Rebased to current `develop` and fixed test failures as a result.

app/decorators/audited/audit_decorator.rb
```diff
@@ -80,5 +82,19 @@ def credit_charge_text(from, to)
     def credit_charge_type
       'usd'
     end
+
+    def tier_level_text(from, to)
+      if to == 3 && from < 3
+        'Escalated this case to tier 3 (General support).'
```
bobwhitelock: We define what the descriptions are for each Tier in the CaseDecorator (https://github.com/alces-software/alces-flight-center/blob/899f214d48e9b5912384ff6e35b11c76db3a336b/app/decorators/case_decorator.rb#L37-L43), so could be worth using this function to avoid duplicating the description in case these ever change.

jamesremuscat:
Rebased on `develop` again and addressed your comment.

bobwhitelock:
LGTM!

PR #247, opened 2018-05-11
Prioritise cases assigned to the current user

atoghill:
This PR implements prioritisation and highlighting of cases that have been assigned to the current user. Cases are split into two distinct groups of those that are and aren't assigned to the current user. With those that are assigned appearing first and foremost at the top in descending order with a darker background to make them stand out.

_(Trello: https://trello.com/c/rvMjwwcA/287-make-cases-assigned-to-current-user-more-prominent-every-time-we-display-case-tables)_

### DIFF
--- a/app/assets/stylesheets/cases.scss
+++ b/app/assets/stylesheets/cases.scss
@@ -45,3 +45,8 @@
 .case-table th {
   text-align: right;
 }
+
+.current-user {
+  background-color: #ecf0f1;
+}
+
--- a/app/assets/stylesheets/cases.scss
+++ b/app/assets/stylesheets/cases.scss
@@ -46,11 +46,6 @@
   text-align: right;
 }
 
-.current-user {
+.assigned-cases {
   background-color: rgba(39,148,216,0.1);
-  #background-color: #ecf0f1;
-}
-
-.current-user td:nth-child(6) {
-  font-weight: bold;
 }
--- a/app/assets/stylesheets/cases.scss
+++ b/app/assets/stylesheets/cases.scss
@@ -47,6 +47,10 @@
 }
 
 .current-user {
-  background-color: #ecf0f1;
+  background-color: rgba(39,148,216,0.1);
+  #background-color: #ecf0f1;
 }
 
+.current-user td:nth-child(6) {
+  font-weight: bold;
+}
--- a/app/views/cases/index.html.erb
+++ b/app/views/cases/index.html.erb
@@ -1,4 +1,3 @@
-<% manage_cases_page = current_user.admin? && @show_resolved %>
 <%= content_for :subtitle do
   "#{@show_resolved ? 'Resolved' : 'Open'} Cases"
 end %>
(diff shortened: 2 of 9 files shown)


bobwhitelock:
I've rebased this on `develop` to (hopefully successfully) reconcile the conflicting changes made both here and on `develop`.

app/views/cases/index.html.erb
```diff
@@ -20,36 +19,15 @@ end %>
...
-              <td style="white-space: nowrap;"><%= c.user.name %></td>
-              <td><%= link_to c.subject, case_path(c) %></td>
-              <td><%= c.assignee&.name || 'Nobody' %></td>
-              <td><%= c.association_info %></td>
-              <% if @show_resolved %>
-                <td><%= c.credit_charge %></td>
-              <% end %>
-            </tr>
-          <% end %>
+        <% @scope.cases.not_assigned_to(current_user).order(created_at: :desc).decorate.each do |c| %>
```
bobwhitelock: It would be nice if we didn't have to explicitly order the Cases by creation date in both sections here - I know we did this before, but since it looks like we do this for Cases in particular in various other places as well:

```
$ git grep 'created_at: :desc'

app/controllers/logs_controller.rb:    @logs = @scope.logs.order(created_at: :desc)
app/controllers/logs_controller.rb:    @cases = @scope.cases.order(created_at: :desc)
app/views/cases/index.html.erb:        <% @scope.cases.assigned_to(current_user).order(created_at: :desc).decorate.each do |c| %>
app/views/cases/index.html.erb:        <% @scope.cases.not_assigned_to(current_user).order(created_at: :desc).decorate.each do |c| %>
app/views/maintenance_windows/_form.html.erb:        maintenance_window.associated_cluster.cases.active.order(created_at: :desc).decorate,
```

and this PR now adds one more, it might be nicer to have this be done automatically everywhere. This could be done by [specifying a default scope](https://stackoverflow.com/questions/3393687/default-sort-order-for-a-rails-model#5609947). Hopefully doing this will just work, though it's possible doing this could cause test failures in some other location if we're expecting Cases somewhere in a different order. If making this change does cause other failures and it's not obvious how to fix, just let me know and we can investigate.

bobwhitelock:
Also: I've rebased this again for you on `develop`, to hopefully remove the unrelated test failures since I've now fixed these.

app/assets/stylesheets/cases.scss
```diff
@@ -45,3 +45,12 @@
 .case-table th {
   text-align: right;
 }
+
+.current-user {
+  background-color: rgba(39,148,216,0.1);
+  #background-color: #ecf0f1;
```
atoghill: Remove this line

app/assets/stylesheets/cases.scss
```diff
@@ -45,3 +45,12 @@
...
+}
+
+.current-user td:nth-child(6) {
+  font-weight: bold;
```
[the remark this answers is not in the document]
atoghill: Opted for `2.` as styling the class would have required another if statement for only adding weight to the text for the current user's cases.

app/views/cases/index.html.erb
atoghill: As of 35330ce these are now ordered by default. However this line is still present in more than one location. If there is a place that would order all such objects by default then please let me know so I can put it there instead.

bobwhitelock: @atoghill you could try ordering everything by date by default in `ApplicationRecord`. Seems like this would normally be what we want to do, though possibly this could cause issues later (if the tests pass then it should be fine for now, and we can always reassess if it later causes any other problems).

atoghill: @bobwhitelock Got a bunch of errors trying to handle it within `ApplicationRecord`, it is highly likely I was doing something wrong but it may be problematic doing it there

app/views/cases/index.html.erb
```diff
@@ -19,14 +19,14 @@ end %>
...
-        <% @scope.cases.assigned_to(current_user).order(created_at: :desc).decorate.each do |c| %>
+      <tbody class="assigned-cases">
+        <% @scope.cases.assigned_to(current_user).decorate.each do |c| %>
           <%= render 'partials/case_table_row', kase: c, show_resolved: @show_resolved %>
         <% end %>
       </tbody>
 
       <tbody>
-        <% @scope.cases.not_assigned_to(current_user).order(created_at: :desc).decorate.each do |c| %>
+        <% @scope.cases.not_assigned_to(current_user).decorate.each do |c| %>
```
bobwhitelock: Not that it's a big deal since you make the rest of the change in the next commit, but removing the `.order` in these two lines should really be in the next commit - with just the changes in this commit the cases could be shown out-of-order here.

bobwhitelock:
All looks great now, thanks @atoghill :shipit:

PR #261, opened 2018-05-14
Admin case management page

atoghill:
The admin dashboard now contains tabs to switch between viewing `All sites` and `All cases`. The former displays what admins currently see on the dashboard. Whereas the latter will display **all** open cases, with the option to view all resolved/closed cases via the dropdown.

Once #247 has been merged then this page should also filter by cases assigned to the current user.

[Trello Card](https://trello.com/c/4V06Dvme/269-add-case-management-pages-for-admins)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -86,11 +86,18 @@ class ApplicationController < ActionController::Base
 
   def assign_site_scope
     if current_user.contact?
+      # For a Site contact, the Site in scope is always the Site they belong
+      # to.
       @site = current_user.site
     elsif request.path =~ /^\/sites/
+      # For an admin viewing the pages for a particular Site, that is the Site
+      # in scope.
       id = scope_id_param(:site_id)
       @site = Site.find(id)
     else
+      # If we reach this point we must be an admin viewing the top-level pages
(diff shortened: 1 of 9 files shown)


app/controllers/application_controller.rb
```diff
@@ -85,19 +85,23 @@ def assign_scope
...
-            end
+    if current_user.contact?
+      @site = current_user.site
+    else
+      if request.path =~ /^\/sites/
+        id = scope_id_param(:site_id)
+        @site = Site.find(id)
+      else
+        AllSites.new
+      end
```
bobwhitelock: I know this is the working solution we ended up with when working on this together yesterday, but looking at this again now I don't see a need for the nested `if`s - the logic would be identical but slightly clearer if we collapse the `else` followed by `if` into `elsif`, and assigning `AllSites.new` would then just be in a top-level `else`.   I think moving the `current_user.contact?` branch first still makes sense, as it means we're handling things for contacts first and then handling the different admin possibilities together.

bobwhitelock:
In the interest of getting this merged I've made those changes, plus a couple of minor things I spotted and some refactoring I've been meaning to do. Thanks for your work on this :shipit:

PR #263, opened 2018-05-15
Improve site dashboard UI/UX

jamesremuscat:
This PR makes some changes to the site dashboard to improve the user experience.

Primarily, it makes more prominent links to each cluster's management dashboard and cases, taking the focus away from the (now renamed) "all site cases" page that was formerly the most obvious.

Site contacts are now listed with clickable email addresses rather than unclickable ones.

I've also taken the opportunity to kill off the Jumbotron page header, replacing it with something a little more compact so that content has a chance of making it "above the fold"; and made some tweaks to the base page layout to reduce the number of empty columns we were rendering.

Trello: https://trello.com/c/bvriA3hT/289-make-site-dashboard-ui-more-intuitive

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -41,6 +41,12 @@
       margin-right: 0.5rem;
     }
   }
+
+  &.title-card {
+    margin-top: 0;
+    margin-bottom: 0;
+    text-align: center;
+  }
 }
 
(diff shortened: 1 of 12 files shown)


jamesremuscat:
Good spot. I blame the sun reflecting off my screen for my not noticing that in the first place ;-)

PR #264, opened 2018-05-15
Add some smoke tests that Case form keeps working

bobwhitelock:
Trello: https://trello.com/c/tPFr5kG4/299-add-some-case-form-feature-tests. I fully expect this to fail in CI, but let's see how this happens.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ rvm:
 sudo: required
 install:
   - bin/setup
+  - rake alces:setup:geckodriver
+# Run tests with xvfb to imitate a display, so Selenium tests work (see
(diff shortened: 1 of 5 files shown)


bobwhitelock:
@jamesremuscat / @atoghill: for reference, you'll need to do one of the steps suggested in https://github.com/alces-software/alces-flight-center/pull/264/commits/ae0399d6aed18c5c7b1bde88b7a5df034bd4359d if you want to run the tests added in this commit locally. Alternatively, you can just use `rspec --tag '~js'` when running tests now, which will just skip these tests (which you may often want to do anyway, since they're slow-ish and pop open Firefox when run).

Issue #233, opened 2018-05-09
Improve maintenance request process


ste78:
When logging a maintenance request directly from the 'Request maintenance' menu item, it makes most sense to log the request against the components that are currently associated with the case, rather than the entire cluster

bobwhitelock:
@ste78 we did do this previously (initially that was the only way to request maintenance IIRC), but switched to the current system of just letting you choose any component/service as people often don't know what the actual root cause of an issue is when they log it. I can see two routes we could go down which would improve the current situation (we could eventually do both if we wanted):

1. As suggested here, only allow requesting maintenance in association with a Case for a Component/Service associated with that Case (or presumably the full Cluster). To make this as flexible as what we allow at the moment we would also need to do the following to support this:

    - allow associating multiple Components and/or Services and/or the whole Cluster with a single Case (https://trello.com/c/x2grQiME/40-want-to-be-able-to-associate-multiple-components-with-single-case) - so can still request maintenance for multiple things in relation to a single Case;

    - allow admins to edit Cases to add Components/Services/the whole Cluster  (https://github.com/alces-software/alces-flight-center/issues/72) - so can still associate maintenance with correct thing following an incorrectly categorised Case or determining what the root cause of a problem was.

2. Make maintenance request process more fully featured - allow selection of the Component/Service to request maintenance for in the maintenance form (similar but simpler than the Case form). The Component(s)/Service(s) associated with the selected Case could be made more prominent when selecting this as this will often be the one you want to select.

We could also do a third stopgap solution to the problem of requesting maintenance for the wrong thing:

3. Indicate more prominently on the maintenance form what you are requesting maintenance for, and indicate how to request maintenance for other things you might want to request maintenance for (e.g. on the Cluster maintenance form say you're requesting maintenance for the Cluster, and link to `/components` and `/services` where you can choose a Component or Service to request maintenance for instead).

bobwhitelock:
Related card on underlying issue: https://trello.com/c/XIf3w3zB/152-make-maintenance-request-process-more-intuitive.

ste78:
Definitely (1). I think the way things are headed cases are going to be at the centre of the whole user experience so I'd shy away from having maintenance requests as independent as in (2). Maybe before cases were such a big part of the system logging the 2 things independently looked to make more sense, but after usage its now pretty clear to me that they are linked. 

Agreed that to do this we'd also need a way to edit case->component/service relationships as well to be sure maintenance requests are logged against correct things.

[...]

PR #265, opened 2018-05-15
Maintenance window request UI enhancements

jamesremuscat:
This PR makes some low-hanging-fruit changes to requesting maintenance windows (MWs) (relating to #233).

- Cases now have a button linking directly to a MW request form for the component/service/cluster associated with that case, and with the appropriate case pre-selected.
- The MW form now tells you what you're about to request maintenance for
- The MW form has a big orange warning if you try and request maintenance for an entire cluster, and suggests that you might want a service or component instead

Trello: https://trello.com/c/XIf3w3zB/290-make-maintenance-request-process-more-intuitive

### DIFF
--- a/app/controllers/maintenance_windows_controller.rb
+++ b/app/controllers/maintenance_windows_controller.rb
@@ -74,6 +74,7 @@ class MaintenanceWindowsController < ApplicationController
 
   def default_maintenance_window_params
     {
+      case_id: params[:case_id],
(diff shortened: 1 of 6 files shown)


jamesremuscat:
Addressed all your points. (Completely didn't realise we used the maintenance form in two different contexts!)

bobwhitelock:
Yep LGTM now, thanks :)

PR #266, opened 2018-05-16
No longer fade out resolved/closed cases

atoghill:
The `Cases` page found under `Resolved` would previously display all cases faded out to indicate they were no longer open. However now that `Open` cases are completely separated from `Resolved/Closed` cases there is no need to fade them out.

This behaviour was only present for users and although not a major issue it did mean the page wasn't very helpful. See [discussion](https://github.com/alces-software/alces-flight-center/pull/237#discussion_r187110712).

[Trello](https://trello.com/c/2DMFKrlZ/293-consider-not-fading-out-every-case-on-cases-resolved-page-for-site-contacts)


app/views/partials/_case_table_row.html.erb
```diff
@@ -6,7 +6,6 @@
 <% if show_resolved != kase.open?  %>
   <% if (kase.resolved? || kase.closed?) && !manage_cases_page %>
     <tr
-      class="closed-case-row"
       title="This support case has been resolved; you should open a new support case if you wish to contact Alces Flight Support"
```
bobwhitelock: Don't think we need this title text as well now that we're no longer fading out the rows, and it's also even less clear now what this is referring to (since it is shown for every row on the resolved Cases page, and is only shown for some columns since sometimes the title text for individual columns takes precedence). Even when you do know what it means it still doesn't make a lot of sense, since even for open Cases there's no longer any way to contact support from the table.

Removing it would also mean we could get rid of this conditional entirely and just use a single `<tr>`, making this template much clearer, so I vote we do this.

atoghill: Yep I agree, do we show this information anywhere else e.g. on the case itself?

bobwhitelock: On the individual Case page (which is where you'd now have to go to leave a comment anyway) we handle whether to allow you to comment and the appropriate message to display if you cannot. So I think you can just remove this whole conditional.

atoghill: 👍 Cool then yeah I'll just remove it completely

PR #267, opened 2018-05-16
Show a more accurate message disabling commenting on non-open cases.

jamesremuscat:
Previously, a resolved or closed tier-2 case would erroneously display a "you can't comment because it's tier 2" message to contacts.

Now, a resolved or closed case of any tier displays a "you can't comment because it's resolved" (or closed) message to all users, and only open tier-2 cases display the former message (still only to contacts).

Trello: https://trello.com/c/kEw2MVZ3/294-should-we-still-show-the-additional-discussion-unavailable-box-for-tier-1-2-cases-when-the-case-is-not-open-and-so-additional-di

### DIFF
--- a/app/decorators/case_decorator.rb
+++ b/app/decorators/case_decorator.rb
@@ -31,6 +31,19 @@ class CaseDecorator < ApplicationDecorator
   end
 
   def commenting_disabled?
-    current_user.contact? && !consultancy?
+    !open? || (current_user.contact? && !consultancy?)
+  end
+
+  def commenting_disabled_text
+    if !open?
+      "Commenting is disabled as this case is #{state}."
+    elsif current_user.contact? && !consultancy?
+      <<~TITLE.squish
+            This is a non-consultancy support case and so additional discussion is
+            not available. If you wish to request additional support please either
+            escalate this case (which may incur a charge), or open a
+            new support case.
+      TITLE
+    end
   end
 end
--- a/app/decorators/case_decorator.rb
+++ b/app/decorators/case_decorator.rb
@@ -31,7 +31,7 @@ class CaseDecorator < ApplicationDecorator
   end
 
   def commenting_disabled?
-    !open? || (current_user.contact? && !consultancy?)
+    commenting_disabled_text.present?
   end
 
   def commenting_disabled_text
--- a/app/views/case_comments/_form.html.erb
+++ b/app/views/case_comments/_form.html.erb
@@ -1,60 +1,28 @@
+<% if @case.commenting_disabled? %>
 
-<%
-  disabled_message_text = <<~TITLE.squish
-            This is a non-consultancy support case and so additional discussion is
-            not available. If you wish to request additional support please open a
-            new support case.
-  TITLE
-  additional_attributes =
-    if @case.commenting_disabled?
-      {
-        disabled: true,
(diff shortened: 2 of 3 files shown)


app/views/case_comments/_form.html.erb
```diff
@@ -17,18 +17,19 @@
```
bobwhitelock: Not directly related to this PR but I just noticed it - we should probably now update the text above to indicate that you can also escalate the Case rather than needing to open a new one if you want to leave a comment, now that this is possible.

app/views/case_comments/_form.html.erb
```diff
@@ -44,4 +45,16 @@
...
+<% else %>
+
+  <div class="card bg-light">
+     <div class="card-body">
+       <span class="float-left mr-3">
+         <%= icon('info') %>
+       </span>
+       Commenting is disabled as this case is <%= @case.state %>.
+     </div>
+  </div>
```
bobwhitelock: With this change we have a few possible/actual inconsistencies in the UI and code:

1. We now have two places in this view we're generating cards for effectively the same purpose, to indicate that commenting is disabled, which probably should continue to remain identical - it would nice if we didn't have to do this twice, and have to remember to update both of these if we want to change how this looks in future.

2. We have a `CaseDecorator#commenting_disabled?` method which can return false even if commenting is actually disabled, which seems confusing.

3. We have two distinct ways of displaying that commenting is disabled (aside from the explanation for why it's disabled, which should definitely remain different) - it can have a comment form with disabled input and button shown, if it is disabled due to the Case's current `tier_level`, or it can have no comment form shown, if it's disabled due to the Case's `status`. Unless there's a good reason for doing these differently I'd suggest we handle these in the same way, and my (not that strong) preference would be for the latter since it's clear what this means and people have already been confused by the large disabled `textarea`.

What are your thoughts on the above? Unless you have a strong preference for keeping the distinction in (3), I'd suggest we move generating the commenting disabled text to a new method in `CaseDecorator`, and then we can simplify this template to something like this:

```ruby
if @case.commenting_disabled?
  # Display the card with the reason why commenting is disabled
else
  # Display the comment form
end
```

jamesremuscat:
Covered off both of your comments in my latest commit.

app/decorators/case_decorator.rb
```diff
@@ -31,6 +31,19 @@ def tier_description
   end
 
   def commenting_disabled?
-    current_user.contact? && !consultancy?
+    !open? || (current_user.contact? && !consultancy?)
```
bobwhitelock: Rather than duplicating the logic I think this method could do something like `!!commenting_disabled_text` or `commenting_disabled_text.present?`, would be clearer and avoid these getting out-of-sync in future.

app/decorators/case_decorator.rb
```diff
@@ -31,6 +31,19 @@ def tier_description
...
+  end
+
+  def commenting_disabled_text
+    if !open?
+      "Commenting is disabled as this case is #{state}."
```
bobwhitelock: Could be worth checking for the presence of this text when testing the form is disabled for this condition.

jamesremuscat:
Done.

bobwhitelock:
Thanks!

PR #268, opened 2018-05-16
Display table headers on a single line

atoghill:
This PR forces the table headers on `Case` pages to display on a single line. Previously they would wrap to a new line to display the second word of `Assigned to` and `Credit usage`. This had the effect of pushing the table content down slightly and as the headers are not massively long seemed unnecessary.

[Trello](https://trello.com/c/8aPaLSua/301-display-table-headers-on-a-single-line)

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -97,11 +97,11 @@ body {
   }
 }
 
-table {
-  .shrink {
-    white-space: nowrap;
-  }
+.nowrap {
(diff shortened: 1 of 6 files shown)


bobwhitelock:
Thanks, looks good - this commit made me notice a change we could make generally which makes this slightly clearer, see the commit I've added for details. I've also then rebased this to resolve the conflicts with my changes on `develop` :neutral_face:. This can be merged once CI passes.

PR #272, opened 2018-05-16
Move cluster documents to their own tab

jamesremuscat:
This PR moves the "Documents" section on the cluster dashboard to its own tab, which is only shown if there are documents present.

Testing for this was a little tricky, so I've had to add 
```ruby
allow_any_instance_of(Cluster).to receive(:documents).and_return([])
```
into the RSpec config in `spec_helper.rb`, overriding it manually in the existing VCR-based test for S3 document retrieval.

For manual testing, I've also dropped a file in S3 for the demo cluster. It's not a long read.

Trello: https://trello.com/c/xomUxdm6/300-move-cluster-documents-into-a-tab

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -141,6 +141,7 @@ Rails.application.routes.draw do
       resources :components, only: :index
       logs.call
       confirm_maintenance_form.call
+      get 'documents', controller: :clusters, action: :documents
     end
 
     resources :components, only: :show do
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -141,7 +141,7 @@ Rails.application.routes.draw do
       resources :components, only: :index
       logs.call
       confirm_maintenance_form.call
-      get 'documents', controller: :clusters, action: :documents
+      get :documents
     end
 
     resources :components, only: :show do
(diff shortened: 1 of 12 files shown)


jamesremuscat:
It's possible I've introduced some nondeterminism in making those changes to the tests. I will investigate...

bobwhitelock:
Bumping the Selenium timeout appears to make the intermittent failures more intermittent (possibly non-existent?). Browser testing :man_shrugging:

config/routes.rb
```diff
@@ -141,6 +141,7 @@
       resources :components, only: :index
       logs.call
       confirm_maintenance_form.call
+      get 'documents', controller: :clusters, action: :documents
```
bobwhitelock: We can use just `get :documents` here and this will work in the same way I believe, as the `controller` and `action` specified here are the defaults.

spec/spec_helper.rb
```diff
@@ -94,4 +94,10 @@
   # test failures related to randomization by passing the same `--seed` value
   # as the one that triggered the failure.
   Kernel.srand config.seed
+
+  config.before :example do
+    # Prevent attempting to retrieve documents from S3 when Cluster page
+    # visited.
+    allow_any_instance_of(Cluster).to receive(:documents).and_return([])
```
bobwhitelock: We have various places in other tests we're still doing the same thing as this - can we remove these now we're stubbing this by default here?

jamesremuscat:
Addressed your comments, plus got rid of the Selenium deprecation warnings we'd introduced.

bobwhitelock:
Thanks!

PR #273, opened 2018-05-17
Show managed and advice based services in cluster overview

atoghill:
This PR adds services to the Cluster Dashboard in the same way that `Components` are listed, with both fully and self managed components.

As the code needed for this was partly shared by the `advice_components` and `managed_components` methods within `app/models/cluster.rb` I extracted them into two new scopes within the `ClusterPart` concern.

This allowed for simple iteration over `Components` and `Services` whilst utilising the shared code to get the number of each respectively from the scopes.
 
[Trello](https://trello.com/c/XFB35TYM/292-show-managed-advice-services-in-cluster-overview)


spec/models/cluster_spec.rb
```diff
@@ -74,7 +74,7 @@
     end
 
     it 'returns all managed cluster components for cluster' do
-      result = subject.managed_components
+      result = subject.components.managed
```
atoghill: @bobwhitelock With the two altered tests I wasn't 100% sure what you'd want to do with them. Currently they are still under the now removed `advice/managed_components` methods although they do still technically test for different types of components being returned. Therefore ` describe #managed_components` is still somewhat relevant. 

However now that we have this functionality under new scopes it is relevant to both `Components` and `Services`.

bobwhitelock: Yeah good point - you could:

- extract these tests to a `shared_examples` file and include these examples in the Component/Service tests;

- change the tests to be dynamic based on the class currently being tested, so they test the behaviour for Components in the Component test file and similarly for Services.

See https://github.com/alces-software/alces-flight-center/blob/feature/show-services-in-cluster-overview/spec/support/shared_examples/inheritable_support_type.rb for some `shared_examples` where we do something quite similar, in particular note the `class_factory_identifier` up the top which lets you dynamically get the factory to use based on the class you're testing.

app/views/clusters/_details.html.erb
```diff
@@ -6,16 +6,19 @@
...
+      <%= part_scope.managed.size %>
+      <%= SupportType::MANAGED_TEXT %>
+      <%= part_type.to_s.pluralize(part_scope.managed.size) %>
+    </li>
+    <li class="list-group-item">
+      <%= part_scope.advice.size %>
+      <%= SupportType::ADVICE_TEXT %>
+      <%= part_type.to_s.pluralize(part_scope.advice.size) %>
+    </li>
+  <% end %>
```
bobwhitelock: Some tests for this section could be useful now it's a bit more complex. Looks like there's no tests for this page in particular atm, so you'd need to add a new feature test file - something like `spec/features/cluster/show_spec.html.erb`. Then within that create a cluster with some managed and advice components and services, and check that the correct text is shown for each.

spec/models/cluster_spec.rb
atoghill: I have extracted the tests to a new shared example and moved the testing to the associated specs for `Component` and `Service`. 

See ca173c9

bobwhitelock:
@atoghill Are you done looking at this again?

atoghill:
Yep this is all done now and ready for review/merging/etc

bobwhitelock:
In which case looks good, thanks :shipit:

PR #274, opened 2018-05-17
Update `flight-chrome`, adding cookie banner.

jamesremuscat:
As title.

Trello: https://trello.com/c/smR8nW1M/297-add-cookie-banner


bobwhitelock:
I tried deleting my cookies and reloading the page and nothing showed - is there something else I need to do to show this?

jamesremuscat:
Clear your cache (or use Chrome's network tools to disable it).

Because we get rid of the hash in the filename of the main script, it stays around longer than we'd like it to.

bobwhitelock:
Thanks, looks great after doing that. One question: does it matter that the same will happen to anyone who has previously visited the site (and so they'll never be shown our cookie warning)?

jamesremuscat:
@bobwhitelock I imagine they'll probably see it "eventually", when their browser decides to check for a newer version.

I'm not sure why it was decided to strip the hash from the script filename originally, but we might want to revisit that decision.

bobwhitelock:
As discussed IRL, looks good and we should fix that at some point

PR #285, opened 2018-05-21
Handle when no items available

atoghill:
This PR doesn't implement anything too fancy but does handle if there are no cases, logs or maintenance to display. In the future this could be adapted to look nicer however for now it does what it needs to. It displays a short message indicating there are no items for the appropriate page to display and prevents the table headers from unnecessarily rendering.

[Trello](https://trello.com/c/W4VOTlw5/46-go-through-and-nicely-handle-when-no-items-available-in-various-cases)

### DIFF
--- a/app/views/cases/index.html.erb
+++ b/app/views/cases/index.html.erb
@@ -1,16 +1,15 @@
 <%
   cases = @show_resolved ? @scope.cases.inactive : @scope.cases.active
+  empty_message = "No #{@show_resolved ? 'resolved or closed' : 'open'} cases"
 %>
 <%= content_for :subtitle do
   "#{cases.length} #{@show_resolved ? 'Resolved' : 'Open'}
   #{'Case'.pluralize(cases.length)}"
 end %>
 <%= render 'partials/tabs', activate: :cases do %>
-  <% if cases.empty? %>
-    <div style="text-align: center">
-      No <%= @show_resolved? 'resolved or closed' : 'open' %> cases
-    </div>
-  <% else %>
+  <%= render 'partials/no_item_available',
+      collection: cases,
+      message: empty_message do %>
     <div class='table-responsive'>
       <table class="table">
         <thead>
--- a/app/views/cases/index.html.erb
+++ b/app/views/cases/index.html.erb
@@ -6,35 +6,41 @@
   #{'Case'.pluralize(cases.length)}"
 end %>
 <%= render 'partials/tabs', activate: :cases do %>
-  <div class='table-responsive'>
-    <table class="table">
-      <thead>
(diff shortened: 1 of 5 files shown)


atoghill:
Rebased on `develop` to get test parity

bobwhitelock:
Rebased on `develop`

app/views/partials/_no_item_available.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if collection.empty? %>
+  <div style="text-align: center"><%= message %></div>
+<% else %>
+  <%= yield if block_given? %>
```
bobwhitelock: I think we always want a block to be passed to this partial, since we always want to render something if the collection is not empty, so I think we should remove the `if block_given?` - this way if a block is not given the code using this partial will not work, which should then be obvious to us and we can fix it (by adding the block we forgot).

app/views/cases/index.html.erb
```diff
@@ -1,40 +1,45 @@
...
-          <% if @show_resolved %>
-            <th>Credit usage</th>
-          <% end %>
-        </tr>
-      </thead>
+  <%= render 'partials/no_item_available',
+      collection: cases,
+      message: empty_message do %>
+    <div class='table-responsive'>
+      <table class="table">
```
bobwhitelock: As discussed the other day, I think we should move these two tags into the partial and where we use the partial have the block just render what is within them. Since these are required for every table it makes sense to not need to duplicate them everywhere within the partial we are using in each table, and instead just do this once, which will also prevent us forgetting this in future tables.

With this change the partial and `message` argument will also need to be renamed, since they will no longer fully reflect what we're doing - possibly to `_table` and `empty_message`. I'm imagining this as the first step towards extracting standard handling for our tables, and we'll then have a place we can pull other things out to as we identify future common things between our tables.

app/views/partials/_no_item_available.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if collection.empty? %>
+  <div style="text-align: center"><%= message %></div>
```
bobwhitelock: I think it would be worth wrapping the `message` in an `<em>` tag, we do this in various places atm to indicate 'this is not the actual data, but just a placeholder where there is no data'.

app/views/cases/index.html.erb
atoghill: Not a big thing but I feel that given the context of the conditional it is clear that `message` is only present when the given `collection` is empty. Renaming it is fine if it makes it cleared but I personally feel it isn't wholly necessary.

bobwhitelock:
LGTM, thanks

PR #286, opened 2018-05-21
Remove legacy RT-integration code

jamesremuscat:
This PR removes now-legacy RT integration code from Flight Center:

- Both real and mocked RT interface classes and tests
- `last_known_ticket_status` from `Case`
- Methods within `Case` designed to update cases from RT tickets
- Validation on RT ticket statuses in `Case`
- VCR configuration and fixtures relating to RT

Note that `rt_ticket_id` remains, as does the logic to display an `RT###` display ID for a case where needed.

I've grepped for "RT" and "deprecated" in `app/` and `spec/` and can't find anything I've missed, though I was expecting this task to be much harder than it turned out...

Trello: https://trello.com/c/YEynY3fS/234-remove-rt-integration-code-mostly-entirely


bobwhitelock:
LGTM, thanks - just removed a few environment variables we had in the `.env.example` file which are now unused. Good that this wasn't too tricky, most RT-related things should have been confined to `RequestTrackerInterface` which we'd mostly removed usage of already.

Can be merged when CI passes.

PR #289, opened 2018-05-21
Primary contact validation

atoghill:
Sites are only meant to have one `Primary Contact` assigned to them however there was no validation to prevent multiple contacts being assigned as the `Primary Contact`. This PR resolves this issue, ensuring that there can be only one. If an admin attempts to assign a user as the `Primary Contact` of a site that already has one then it will raise an error within the admin interface preventing the user from being saved.

[Trello](https://trello.com/c/NQZqLwPF/110-add-validation-that-site-only-has-single-primary-contact)

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -20,7 +20,7 @@ RSpec.describe User, type: :model do
 
   describe '#validates_primary_contact_assignment' do
     subject do
-      create(:user, primary_contact: true)
+      build(:user, primary_contact: true)
     end
 
     context 'with no primary contact' do
@@ -30,10 +30,10 @@ RSpec.describe User, type: :model do
     end
 
     context 'with an existing primary contact for the current site' do
-      let :primary_contact { create(:user, primary_contact: true, site: subject.site) }
+      let! :primary_contact { create(:user, primary_contact: true, site: subject.site) }
 
       it 'should error' do
-        expect{ primary_contact }.to raise_error(ActiveRecord::RecordInvalid)
+        expect(subject).not_to be_valid
       end
     end
   end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -17,4 +17,24 @@ RSpec.describe User, type: :model do
       it { is_expected.to be false }
     end
   end
+
+  describe '#validates_primary_contact_assignment' do
+    subject do
+      create(:user, primary_contact: true)
+    end
+
+    context 'with no primary contact' do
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -17,4 +17,24 @@
...
+      it 'should not error' do
+        expect(subject).to be_valid
+      end
+    end
+
+    context 'with an existing primary contact for the current site' do
+      let! :primary_contact { create(:user, primary_contact: true, site: subject.site) }
+
+      it 'should error' do
+        expect(subject).not_to be_valid
```
bobwhitelock: If an object is valid, every validation it has must have passed, so it is sufficient to check that it is valid (as you do above) to check that we're not incorrectly invalidating it.

If it is invalid, though, there are many reasons why this could be the case, including ones entirely incidental to the thing you are testing - therefore I think as well as this line, checking it is not valid, it is also worth adding an assertion that the error message you expect is present, e.g. like https://github.com/alces-software/alces-flight-center/blob/60282bc1c61683290ba8ffc5811a201e7432162c/spec/models/issue_spec.rb#L16-L18

This should ensure the model is invalid for the reason you expect, with the error you expect, rather than it being possible for it to be invalid in future for reasons unrelated to what you are testing.

atoghill: Just want to point out that I initially had something like this in my test and you told me to take it out. I'm more than happy to add it back in but the mixed messages aren't exactly helpful

bobwhitelock: I don't think I did or would have, possibly we miscommunicated as I was discussing other ways to improve the test. Sorry if I wasn't very clear at the time :slightly_smiling_face:. I'd always suggest testing that the error message you expect is present, as there are many, sometimes surprising, ways your objects can become invalid.

atoghill: No you 100% did, mind you this was in the time where we were waiting for the meeting so you may have been distracted. You told me that it was bad to test the string as if the string ever changed then the test would fail.

bobwhitelock: Hard to remember without the original code in front of me, and sorry that I wasn't clearer at the time - I was slightly distracted as was trying to fix our staging deploy as well. If you have the original code I can try to clarify what I must have meant.

atoghill: It's not a big deal, pretty sure that original code is long gone now. But as of 3d1c1a4 it is being tested again.

bobwhitelock:
Thanks

PR #293, opened 2018-05-22
Validate services for advice based clusters

atoghill:
This PR validates `Services` in the same manner that `Components` are validated for advice based clusters.

[Trello](https://trello.com/c/AK9Truy4/120-missing-corresponding-validation-that-advice-cluster-does-not-have-managed-services-as-for-components)

### DIFF
--- a/app/models/cluster.rb
+++ b/app/models/cluster.rb
@@ -18,8 +18,7 @@ class Cluster < ApplicationRecord
   validates :name, presence: true
   validates :support_type, inclusion: { in: SUPPORT_TYPES }, presence: true
   validates :canonical_name, presence: true
-  validate :validate_all_components_advice, if: :advice?
-  validate :validate_all_services_advice, if: :advice?
+  validate :validate_all_cluster_parts_advice, if: :advice?
   validates :shortcode, presence: true, uniqueness: true
   # validates_presence_of :motd
 
@@ -76,17 +75,14 @@ class Cluster < ApplicationRecord
 
   private
 
-  def validate_all_components_advice
-    unless components.all?(&:advice?)
-      errors.add(:base, 'advice Cluster cannot be associated with managed Components')
+  def validate_all_cluster_parts_advice
+    ['components', 'services'].each do |cluster_part|
+      unless eval(cluster_part).all?(&:advice?)
+        errors.add(:base, "advice Cluster cannot be associated with managed #{cluster_part.capitalize}")
+      end
     end
   end
 
-  def validate_all_services_advice
-    unless services.all?(&:advice?)
-      errors.add(:base, 'advice Cluster cannot be associated with managed Services')
-    end
-  end
 
   def create_automatic_services
     ServiceType.automatic.each do |service_type|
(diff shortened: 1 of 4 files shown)


app/models/cluster.rb
```diff
@@ -75,12 +75,15 @@ def next_case_index
 
   private
 
-  def validate_all_components_advice
-    unless components.all?(&:advice?)
-      errors.add(:base, 'advice Cluster cannot be associated with managed Components')
+  def validate_all_cluster_parts_advice
+    ['components', 'services'].each do |cluster_part|
+      unless eval(cluster_part).all?(&:advice?)
```
bobwhitelock: I know we can be pretty sure what we're `eval`ing here, as (at least for now) we're defining it on the line above, but this use of `eval` still makes me nervous. In general you almost always don't need `eval`, and if you can avoid it you shouldn't use it (see https://stackoverflow.com/questions/1902744/when-is-eval-in-ruby-justified).

In this case I believe you could use `self.public_send(cluster_part).all?(&:advice)` and achieve the same effect, with much less scope for things going terribly wrong if something changes in future; using `public_send` for this is also more standard and less likely to confuse/worry readers of this method in future.

app/models/cluster.rb
```diff
@@ -75,12 +75,15 @@ def next_case_index
...
+        errors.add(:base, "advice Cluster cannot be associated with managed #{cluster_part.capitalize}")
+      end
     end
   end
 
+
```
bobwhitelock: Pedantry: why the extra whitespace?

atoghill: Not there intentionally, cheers for finding that 👍

app/models/cluster.rb
atoghill: Yep I agree with that now, also let me know if this all works as intended. The tests are passing but I'm still not 100% sure everything is fine.

bobwhitelock:
@atoghill Thanks for making the change, looks good - think I've spotted the cause of what you mentioned on Slack also: you said when you created a managed service for an advice cluster, this was allowed and it was not indicated that this was invalid. Reason for this I think is that these validations live on the Cluster, and are not run when creating other objects (like Components and Services). We should be able to fix this by adding `validates_associated :cluster` to both Component and Service, so that if creating a Component/Service invalidates the Cluster this will fail the validation when you attempt to do this.

atoghill:
@bobwhitelock Ah that sounds logical. Does that mean the validation for `Components` wasn't really being handled before?

bobwhitelock:
@atoghill it means it was being handled when a cluster is validated, since it's a validation on Cluster, and so you wouldn't be able to change a Cluster to `advice` if it had any explicitly `managed` Components. It wasn't being run though when you change a Component (and now Service), but since that can indirectly invalidate a Cluster we should probably also run it in this case.

atoghill:
@bobwhitelock Alright cool, should be fixed in 95c032a

bobwhitelock:
LGTM, thanks

PR #308, opened 2018-05-24
Document development process

bobwhitelock:
This PR adds documentation to the README for setting up a development environment and common development tasks. Some of this stuff should be obvious/standard for a Rails app but thought I might as well make things explicit.

@jamesremuscat and @atoghill would you be able to check to see if anything looks unclear/missing/inaccurate to you?

@benarmston barring any needed changes, this should document what you'll need to do to set things up and to do some common development tasks. Let me know and/or fix if anything here doesn't work for you or make sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
     ```bash
     git clone git@github.com:alces-software/alces-flight-center.git
     cd alces-flight-center
-    bin/setup
+    bundle exec bin/setup
     ```
 
 6. Perform required/suggested additional manual setup steps output at the end
--- a/README.md
+++ b/README.md
@@ -1,4 +1,136 @@
-# README
+
+# Alces Flight Center
+
+## Development
+
+### To setup a development environment
+
+1. Have Ruby version specified in [`.ruby-version`](./.ruby-version) installed
+   (if you don't have a preferred method of installing Ruby, I recommend using
+   [`rbenv`](https://github.com/rbenv/rbenv) and
+   [`ruby-build`](https://github.com/rbenv/ruby-build)).
+
+2. Have PostgreSQL version specified in [`.travis.yml`](./.travis.yml)
+   installed and running, and configured so you can connect to it as the
+   `postgres` user without a password. On CentOS the following should achieve
+   this step (other OSes may vary):
+
+    ```bash
+    yum install postgresql96 postgresql96-devel postgresql96-server
+    postgresql-setup initdb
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,4 +1,136 @@
...
+   [`rbenv`](https://github.com/rbenv/rbenv) and
+   [`ruby-build`](https://github.com/rbenv/ruby-build)).
+
+2. Have PostgreSQL version specified in [`.travis.yml`](./.travis.yml)
+   installed and running, and configured so you can connect to it as the
+   `postgres` user without a password. On CentOS the following should achieve
+   this step (other OSes may vary):
+
+    ```bash
+    yum install postgresql96 postgresql96-devel postgresql96-server
```
atoghill: Not sure it is necessary to install `postgresql96`  any more. The other day when you upgraded `Ruby` I needed to install `postgresql-devel` to get things functional again. Might need to investigate this further?

bobwhitelock: Yeah I wasn't sure if I should include this bit, but it documents the steps we needed to do to set Postgres up correctly in one particular case (on your CentOS VM), so I thought it could still be useful in future as a reference.

To expand on the bit above this though, the general requirement for this is that you need Postgres, of a version with support for JSON columns (which might as well be 9.6 since that's what we use in production atm, to reduce other possible issues due to dev/prod differences), and configured so you can sign in without a password as the `postgres` user (which probably could be made a less strict requirement, but that's what our current `config/database.yml` expects).

If you think we don't need all these steps/need additional steps in your case then would be great if you could check this (e.g. by `yum remove`ing things until things no longer work) and let me know/update this section. But it's likely not a big deal if you don't want to check this, I think you are/will be the only one running Flight Center with anything Red Hat-based so this is just a rough guideline of what might be needed for other OSes. In general I imagine this step will vary a lot based on however you're setting up Postgres, which will depend on your particular machine and OS.

atoghill: Okay after more thorough investigation I believe these steps are still necessary. Unsure exactly why I was prompted to install `postgresql-devel` which is version `9.2.23`. If anyone runs into problems with this in the future it should be fairly obvious what they need to do anyway :)

benarmston:
@bobwhitelock these steps have worked for me.  I've changed `bin/setup` to `bundle exec bin/setup` as that was required for me.  I presume that its required/acceptable to everyone else's environments too.

bobwhitelock:
Going to merge this, since @atoghill and @benarmston say it's fine and @jamesremuscat is no longer here.

PR #316, opened 2018-05-24
Update Flight Chrome

benarmston:
@bobwhitelock this PR adds 3 links to the footer: "Privacy policy", "Terms" and "Security".  Those links redirect to the appropriate pages on the main website.  As those pages are generic, this should be sufficient for flight center use.


benarmston:
Thanks for your comments on 3ba049c @bobwhitelock. I think I have a better understanding of how this is supposed to hang together now.

Issue #232, opened 2018-05-09
Cluster Wiki


ste78:
Wiki / markdown pages for a cluster, admin sections for our notes on how to manage various aspects of the cluster and an user facing version for customers own notes.

bobwhitelock:
Trello: https://trello.com/c/Rp6uxL44/172-cluster-wiki

bobwhitelock:
Was done in above PR.

PR #334, opened 2018-06-05
Cluster notes

benarmston:
This PR implements #232.  Currently, a cluster can have `engineering` and `customer` notes.  Admin users can view, create and update both kinds of notes, site contacts can view, create and update only `customer` notes.

@bobwhitelock I have a few questions:

 1. Is this along the lines of what you were expecting?
 2. I'm assuming that we want to add the ability for the `customer` notes to be edited created through the non-RailsAdmin UI.  Is that correct?
 3. Do we use the RailsAdmin UI in production for creating/editing/deleting admin content?
 4. Any other comments you have would be appreciated.

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -37,13 +37,8 @@ class NotesController < ApplicationController
   private
 
   def note_from_params
-    cluster = cluster_from_params
     flavour = params.require(:flavour)
-    cluster.notes.send(flavour).first || cluster.notes.new(flavour: flavour)
-  end
-
-  def cluster_from_params
-    Cluster.find(params.require(:cluster_id))
+    @cluster.notes.send(flavour).first || @cluster.notes.new(flavour: flavour)
   end
 
   def note_params
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -3,11 +3,7 @@ class NotesController < ApplicationController
 
   def show
     @note = note_from_params
-    if @note.persisted?
-      render :show
-    else
-      render :new
-    end
+    render :new unless @note.persisted?
   end
 
   def edit
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -20,7 +20,7 @@ class NotesController < ApplicationController
   end
 
   def edit
-    @note = Note.find(params[:id])
+    @note = note_from_params
   end
 
   def create
@@ -29,28 +29,32 @@ class NotesController < ApplicationController
     if @note.save
       flash[:success] = 'Notes created'
     else
-      error_flash_models [@note], "Your notes were not created. #{@note.errors.full_messages.join('; ').strip}"
+      flash[:error] = "Your notes were not created: #{format_errors(@note)}"
     end
     redirect_back fallback_location: cluster
   end
 
   def update
-    @note = Note.find(params[:id])
+    @note = note_from_params
     if @note.update_attributes(note_params)
       flash[:success] = 'Notes updated'
       redirect_to send("#{@note.flavour}_cluster_notes_path", @note.cluster)
     else
-      error_flash_models [@note], "Your notes were not updated. #{@note.errors.full_messages.join('; ').strip}"
+      flash[:error] = "Your notes were not updated: #{format_errors(@note)}"
       render :edit
     end
   end
 
   private
 
-  def note_from_params(flavour)
-    cluster = cluster_from_params
-    note = cluster.notes.send(flavour).first || cluster.notes.new(flavour: flavour)
-    @note = note.decorate
+  def note_from_params(flavour=nil)
+    if params[:id]
+      note = Note.find(params[:id])
+    else 
+      cluster = cluster_from_params
+      note = cluster.notes.send(flavour).first || cluster.notes.new(flavour: flavour)
+    end
+    note.decorate
   end
 
   def cluster_from_params
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -3,19 +3,61 @@ class NotesController < ApplicationController
 
   def engineering
     @note = note_from_params(:engineering)
-    render :show
+    if @note.persisted?
+      render :show
+    else
+      render :new
+    end
   end
 
   def customer
     @note = note_from_params(:customer)
-    render :show
+    if @note.persisted?
+      render :show
+    else
+      render :new
+    end
+  end
+
+  def edit
+    @note = Note.find(params[:id])
+  end
+
+  def create
+    cluster = cluster_from_params
+    @note = cluster.notes.build(note_params)
+    if @note.save
+      flash[:success] = 'Notes created'
+    else
+      error_flash_models [@note], "Your notes were not created. #{@note.errors.full_messages.join('; ').strip}"
+    end
+    redirect_back fallback_location: cluster
+  end
+
+  def update
+    @note = Note.find(params[:id])
+    if @note.update_attributes(note_params)
+      flash[:success] = 'Notes updated'
+      redirect_to send("#{@note.flavour}_cluster_notes_path", @note.cluster)
+    else
+      error_flash_models [@note], "Your notes were not updated. #{@note.errors.full_messages.join('; ').strip}"
+      render :edit
+    end
   end
 
   private
 
   def note_from_params(flavour)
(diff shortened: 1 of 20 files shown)


app/decorators/note_decorator.rb
```diff
@@ -0,0 +1,21 @@
...
+    else
+      "Notes"
+    end
+  end
+
+  def not_found_message
+    if current_user.admin?
+      "No #{flavour} notes have been addded yet."
+    else
+      "No notes have been addded yet."
```
bobwhitelock: FYI `s/addded/added/` here and above :slightly_smiling_face:

app/models/concerns/admin_config/note.rb
```diff
@@ -0,0 +1,20 @@
...
+    rails_admin do
+      edit do
+        configure :description do
+          html_attributes rows: 10, cols: 100
+          help Constants::MARKDOWN_DESCRIPTION_EDIT_HELP
+        end
+
+        configure :flavour do
+          help <<~EOF
+            Valid flavours are #{Note::FLAVOURS.join(', ')}.
```
bobwhitelock: FWIW if you define a `Note#flavour_enum` method this will be automatically picked up by `rails-admin` and only these flavours will be selectable, then you won't need this help :slightly_smiling_face:. See here for where we do something similar: https://github.com/alces-software/alces-flight-center/blob/a3c31dadf894ea044def036e5809901a0bbf6f6c/app/models/cluster.rb#L39-L43

benarmston:
@bobwhitelock thanks for the comments above.  I've addressed the issues you mentioned, implemented, create/update support and written some tests.  Additional features that we're likely going to want at some point might include:

 - Preview support, similar to the github comment box.
 - A rich text editor, similar to the github comment box.

I can look at implementing one or both of those, or move onto something else.

app/controllers/notes_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def update
+    @note = note_from_params
+    authorize @note
+    if @note.update_attributes(note_params)
+      flash[:success] = 'Notes updated'
+      redirect_to @note.decorate.path
+    else
+      flash[:error] = "Your notes were not updated: #{format_errors(@note)}"
```
bobwhitelock: If I create some notes, then edit them and attempt to delete all content, I get an error as `Your notes were not updated: description can't be blank` - however I can imagine valid use cases where a User may want to do this, e.g.:

1. a User adds a note to a Cluster, which previously had no notes, about some temporary issue to be aware of related to the Cluster;

2. later this issue no longer exists so they want to delete everything in the note;

3. currently to do this they would need to still enter some placeholder text so that the existing note can be saved.

Possibly we should either allow `description` to be empty when updating, or `destroy` the existing note when `update` is submitted with a blank description (this would also have the advantage of re-displaying the 'no notes yet' message in this situation). What do you think to this?

app/controllers/notes_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+      render :edit
+    end
+  end
+
+  private
+
+  def note_from_params
+    cluster = cluster_from_params
```
bobwhitelock: We have some code in `ApplicationController` to find the top-level object that the User is on the dashboard for (which we call the scope): https://github.com/alces-software/alces-flight-center/blob/ee18cddfdbd57440550b80b51284cc21e2516405/app/controllers/application_controller.rb#L69-L88

So I think you could make use of this by using `@cluster` here, which would avoid `find`ing the Cluster twice.

(I realise this is not that discoverable atm, and there's probably some room for improvement with this stuff!)

app/views/notes/new.html.erb
```diff
@@ -0,0 +1,17 @@
...
+        <%= form_for note.form_path do |f| %>
+            <div class="form-group">
+                <%= f.text_area :description, class: 'form-control', rows: 10 %>
+            </div>
+            <div class="form-group" style="overflow: hidden;">
+                <%= f.submit "Create notes", class: dark_button_classes %>
+            </div>
+        <% end %>
+    </div>
+<% end %>
```
bobwhitelock: The `new` and `edit` views currently look very similar - possibly worth extracting a partial for these which could just be passed the text differences? Would avoid them getting out-of-sync and/or us needing to remember to update both together.

app/views/notes/new.html.erb
```diff
@@ -0,0 +1,17 @@
+<% content_for(:subtitle) { note.subtitle } %>
+
+<%= render 'partials/tabs', activate: :notes do %>
+    <div class="card-body">
+        <p>
+        <%= note.new_form_intro %>
+        </p>
+        <%= form_for note.form_path do |f| %>
+            <div class="form-group">
+                <%= f.text_area :description, class: 'form-control', rows: 10 %>
```
bobwhitelock: Possibly worth indicating that this description will be rendered as markdown, as customers may not be aware of this (same for the `edit` form).

app/controllers/notes_controller.rb
```diff
@@ -0,0 +1,56 @@
+class NotesController < ApplicationController
+  decorates_assigned :note
+
+  def show
+    @note = note_from_params
+    if @note.persisted?
+      render :show
+    else
+      render :new
+    end
```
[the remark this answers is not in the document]
benarmston: I've not much preference.  I'll change it as you suggest.

app/controllers/notes_controller.rb
benarmston: Let's destroy the note when it has no description.

app/controllers/notes_controller.rb
benarmston: OK. I'll swap out to using the auto assigned `@cluster`.  Thanks for pointing me at this.

Regarding discoverability, I think that's going to be an issue for anyone joining any largeish project.

app/views/notes/new.html.erb
benarmston: OK. I'll do that.

app/views/notes/new.html.erb
benarmston: I'll add a very simple indication of that.  I'll leave link to the markdown flavour we support for a future enhancement.

benarmston:
@bobwhitelock I've addressed all of your comments.

app/controllers/notes_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    redirect_back fallback_location: @note.cluster
+  end
+
+  def update
+    @note = note_from_params
+    authorize @note
+    if note_params[:description].blank?
+      @note.destroy
+      flash[:success] = 'Notes removed'
+      redirect_to cluster_path(@note.cluster)
```
bobwhitelock: Do we not want to `redirect_to @note.decorate.path` in this case too? Being redirected back to the cluster overview page seems slightly surprising.

benarmston: I don't think so.  If we did, we'd have the following interaction:

 - user: visits the note page
 - user: decides to remove note
 - centre: informs user (via flash) that the removal was successful
 - centre: asks user if they would like to create the note that they just deleted.

I can't imagine the use case in which that would leave the user on the correct page.

bobwhitelock: I guess the same is true with the language suggesting 'No notes have been added for this cluster yet' if they then navigate back to the notes page, which seems surprising in the same way since notes have been added (and then later removed) for the cluster - so possibly the wording should be changed to not imply this, to something like 'there are currently no $flavour notes for this cluster'.

I was thinking that, from a users perspective, their intention when deleting all the content in an existing note is to just have that note be empty, and the fact that behind the scenes this actually deletes the model is somewhat an implementation detail. Therefore it could surprise people that if they leave a single character in the description they will be redirected back to the same page, but if they delete one more character they will be redirected to a different page, even though the original page is still a valid URL they can visit (and they can immediately navigate back here if they want).

In any case maybe real Users will think differently to me, so happy for you to tweak the wording/change the redirect and/or just merge this as you prefer, we can always adjust in response to user feedback :slightly_smiling_face:.

benarmston: I like the changed wording. I'll make that change.

benarmston: If I understand your last paragraph correctly, you are happy for me to merge this PR to develop.  Is that correct?

bobwhitelock: Yep, fine by me :slightly_smiling_face:

PR #342, opened 2018-06-07
Markdown all the things

benarmston:
This PR adds support for markdown in the log `details` and case comment `text` columns.

A `MarkdownColumn` "module" has been added to support markdown descriptions in columns other than `description`.  An example of how to configure the column to use is show below:

```ruby
class Log < ApplicationRecord
  include MarkdownColumn(:details)
end
```

In the above example, `Log` will have a `rendered_details` method, which returns the result of rendering the content of the `details` column as markdown.

The `MarkdownDescription` module is now implemented in terms of `MarkdownColumn` and has been kept for backwards compatibility.

Markdown rendering has been changed to strip out all block level html attributes.  This has been done to prevent `<script>alert('I hakorz you')</script>` attacks.

Finally, the `logs/index.html.erb` template has been updated to use the markdown rendered details, and similarly for the  case comments and logs case events.

---

Logs and case comments is a small value of "all the things" so please let me know if there is anything else that should be markdownified.


benarmston:
Still todo: tests

benarmston:
We now have tests.  When #334 has been merged I'll rebase / edit this PR to make the duplicate commits go away.

benarmston:
@bobwhitelock this has now been rebased on the latest develop.  This is ready for review and that should be much easier than it would have been earlier.

app/decorators/case_comment_decorator.rb
```diff
@@ -4,7 +4,8 @@ def event_card
     h.render 'cases/event',
              name: object.user.name,
              date: object.created_at,
-             text: object.text,
+             text: object.rendered_text.html_safe,
```
bobwhitelock: Would it be possible/make sense to call `html_safe` within the `rendered_*` methods themselves? Since if we (almost) always want to do it when using the rendered markdown, and there's no negative consequences to doing it always (since the rendered markdown is HTML-safe, as we appropriately handle user input and strip `<script>` tags etc.), this could prevent us from forgetting to do this anywhere we want to insert the rendered markdown.

benarmston: Good idea.  I've made the change. I'll merge the PR when the checks have passed.

bobwhitelock:
:shipit:

PR #345, opened 2018-06-11
Update dependencies with `bundle update`.

jamesremuscat:
This was motivated by a Github security advisory for an XSS vulnerability in the version of Sinatra we were pulling in (as a dependency of Resque).

Trello: https://trello.com/c/Q9V0OiCo/347-update-sinatra-gem-to-avoid-xss-vulnerability


bobwhitelock:
LGTM, thanks

PR #346, opened 2018-06-11
motd change request button

benarmston:
This PR adds a button on the cluster's overview in the MOTD section to display the case form ready populated for creating a "MOTD change request".  

This works by adding `data-selected-category` and `data-seleted-issue` to the `new-case-form` DOM element which are passed into the elm app along with the other flags.  The elm `init` function then returns a list of `Cmd Msg`s to set the desired category and issue.

Whist this works, it would be better to have `decodeInitialModel` set the desired category and issue from `flags`.  Doing that would avoid a couple of trips through the elm runtime.  Unfortunately, how to do that is not immediately obvious to me.

Still to do: ~tests~

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -39,6 +39,15 @@ class CasesController < ApplicationController
                 else
                   current_site.clusters
                 end
+    @selected = if params[:tool].present?
+                  tool = Tier.find_by(tool: params[:tool])
+                  {
+                    issue: tool.issue_id,
+                    category: tool.issue.category_id
+                  }
+                else
+                  {}
+                end
   end
 
   def create
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -49,7 +49,7 @@ class CasesController < ApplicationController
                     {
                       category: issue.category_id,
                       issue: issue.id,
-                      service: service.id,
+                      service: service.present? ? service.id : nil,
                       tier: tool.level,
                     }
                   end
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -41,12 +41,18 @@ class CasesController < ApplicationController
                 end
     @selected = if params[:tool].present?
                   tool = Tier.find_by(tool: params[:tool])
-                  tool.nil? ?
-                    {} :
+                  if tool.nil?
+                    {}
+                  else
+                    issue = tool.issue
+                    service = issue.service_type.services.find_by(cluster_id: cluster_id)
                     {
-                      issue: tool.issue_id,
-                      category: tool.issue.category_id
+                      category: issue.category_id,
+                      issue: issue.id,
+                      service: service.id,
+                      tier: tool.level,
                     }
+                  end
                 else
                   {}
                 end
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -39,24 +39,7 @@ class CasesController < ApplicationController
                 else
                   current_site.clusters
                 end
-    @selected = if params[:tool].present?
-                  tool = Tier.find_by(tool: params[:tool])
-                  if tool.nil?
-                    {}
-                  else
-                    # issue = tool.issue
-                    # service = issue.service_type.services.find_by(cluster_id: cluster_id)
-                    {
-                      # category: issue.category_id,
-                      # issue: issue.id,
-                      # service: service.present? ? service.id : nil,
-                      # tier: tool.level,
-                      tool: params[:tool],
-                    }
-                  end
-                else
-                  {}
-                end
+    @pre_selected = get_pre_selections
   end
 
   def create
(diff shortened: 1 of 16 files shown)


benarmston:
@bobwhitelock whilst writing the tests for this I discovered that some of the issue factories do not have an associated category.  Could you clarify when issues should have categories please?

spec/features/case/new_spec.rb
```diff
@@ -88,6 +88,7 @@
 
       visit new_case_path(as: user)
       fill_in 'New MOTD', with: motd
+      sleep 20
```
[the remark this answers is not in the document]
benarmston: We don't.  I added it to better understand how the fixtures were working for that test. I thought I'd removed it before committing.  I'll do that now.

app/controllers/cases_controller.rb
```diff
@@ -35,6 +35,17 @@ def new
...
+    @selected = if params[:tool].present?
+                  tool = Tier.find_by(tool: params[:tool])
+                  tool.nil? ?
+                    {} :
+                    {
+                      issue: tool.issue_id,
+                      category: tool.issue.category_id
+                    }
+                else
+                  @selected = {}
```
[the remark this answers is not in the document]
benarmston: Good spot.

bobwhitelock:
Not sure if you've still got anything you want to add to this @benarmston, but if not happy for this to be merged as it is :)

benarmston:
I've decided implementing your 3rd suggested fix will be a good way of familiarising myself with elm.  If that goes well I'll have other things for you to look at.  If not, I'll bail on that and merge.  I'll let you know either way.

bobwhitelock:
OK sounds good, good luck :slightly_smiling_face:

benarmston:
@bobwhitelock I've now implemented your suggested solutions 2 and 3 above.  Doing so has given me a much better understanding of Elm and of the case form.  If you have any comments on the case form changes, I'd love to hear them.

As I had already implemented support for selecting a specific service, category, issue and tier, I have left in support for that.  Though I'm not sure that doing so is such a good idea.

The code adds a some complexity, that isn't needed to support selecting a tier 1 tool.  As the card I was working on only required support for selecting a tool.  It would perhaps be better to remove the code.  However, before I do so, I thought I'd mention it in case you know of any pending requirement for this.

The service, category, issue selecting code works mostly as expected, but does have some edge cases with issues that don't have a support type, such as "Suspected service issue".  If it worked without edge cases, I'd be more inclined to leave it.

app/javascript/packs/Service.elm
```diff
@@ -67,3 +68,75 @@ setSelectedCategory : Category.Id -> Service -> Service
...
+                        _ ->
+                            Nothing
+
+                Issues.JustIssues issues ->
+                    case findIssueAndTier predicate issues of
+                        Just ( issue, tier ) ->
+                            Just ( service, Nothing, issue, tier )
+
+                        _ ->
+                            Nothing
```
bobwhitelock: In the different `case`s of `Maybe`s in this file you might be able to simplify things slightly using `Maybe.map` - would avoid you needing to unpack and then recreate the `Maybe`s yourself.

app/javascript/packs/State/Update.elm
```diff
@@ -158,6 +161,48 @@ handleChangeTierField state index value =
...
 
 
+handleSelectTool : State -> String -> ( State, Cmd Msg )
+handleSelectTool state toolName =
+    let
+        services =
+            state
+                |> .clusters
+                |> SelectList.selected
+                |> .services
```
bobwhitelock: This could be slightly simplified to `State.selectedCluster state |> .services` I believe.

app/javascript/packs/State/Update.elm
```diff
@@ -158,6 +161,48 @@ handleChangeTierField state index value =
...
+                    state.clusters
+                        |> flip Cluster.setSelectedService service.id
+                        |> maybeSetCategory mcategory
+                        |> flip Cluster.setSelectedIssue (issue |> Issue.data |> .id)
+                        |> flip Cluster.setSelectedTier tier.id
+            }
+                ! []
+
+        Nothing ->
+            state ! []
```
bobwhitelock: Could use `Maybe.map` here I think so we only need to consider the stuff inside the `Maybe`.

app/javascript/packs/State/Update.elm
```diff
@@ -158,6 +161,48 @@ handleChangeTierField state index value =
...
+                    clusters
+    in
+    case Service.findTierAndAncestors isWantedTier services of
+        Just ( service, mcategory, issue, tier ) ->
+            { state
+                | clusters =
+                    state.clusters
+                        |> flip Cluster.setSelectedService service.id
+                        |> maybeSetCategory mcategory
+                        |> flip Cluster.setSelectedIssue (issue |> Issue.data |> .id)
```
bobwhitelock: A minor tweak, but could be worth pulling out an `Issue.id` method for extracting the `id` for symmetry with the other methods for unpacking Issue `data` (like `Issue.name` etc.)

app/javascript/packs/Service.elm
```diff
@@ -67,3 +68,75 @@ setSelectedCategory : Category.Id -> Service -> Service
...
+    in
+    services
+        |> SelectList.map findServiceCategoryIssueAndTier
+        |> SelectList.Extra.find Maybe.Extra.isJust
+        |> Maybe.Extra.join
```
bobwhitelock: We do this same pattern in 3 places (`SelectList.map ... |> SelectList.Extra.find Maybe.Extra.isJust |> Maybe.Extra.join`), so could make things slightly simpler to extract a shared function for this.

spec/factories/tier.rb
```diff
@@ -13,7 +13,8 @@
       level 1
 
       factory :tier_with_tool do
-        tool :motd
+        association :issue, factory: :issue_with_category
+        tool 'motd'
```
bobwhitelock: Not that I mind the change just curious - was something breaking with using a symbol here? Will avoid using symbols in similar cases in future if so :slightly_smiling_face:

app/javascript/packs/Service.elm
benarmston: There is one point in that file where I have

```
                findTierAddingIssue issue =
                    Maybe.map ((,) issue) (Issue.findTier predicate issue)

```

but I couldn't find a general way of going from an n-tuple to an n+1-tuple.  Unless you know of the function that I need, I'm going to leave this as it is.

app/javascript/packs/State/Update.elm
benarmston: Excellent. Thank you.

app/javascript/packs/State/Update.elm
benarmston: Good idea.

app/javascript/packs/Service.elm
benarmston: Good idea.  We now have `SelectList.Extra.findValueBy`.  Not sure that's the best name, but it'll do for now.

spec/factories/tier.rb
benarmston: One of my tests was matching against that value.  I had `"motd" == :motd` causing the test to fail.  This change made the test work.  But on further thought I think the test would be more robust if I changed the test to call `.to_s`.  I'll make that change.

benarmston: Or perhaps it was an entirely unnecessary change ;-)

app/javascript/packs/State/Update.elm
benarmston: Getting the types to line up with this is going to take a bit too much effort. So I'm going to leave it as it is.

PR #347, opened 2018-06-11
Allow enabling contact comments on tier 1/2 cases

### LINKED ISSUE #319: Allow admins to enable / disable user communication box on a tier 1/2 case
Problem:

- User logs tier 2 issues
- Admin needs more information and ask the user a question
- User needs to be able to respond but keep the case as a tier 2

Suggested Workflow:

- Customer logs tier 2 case
- Admin confirms a valid tier 2 issue, asks the question and elects to enable 2 way communication
- User can see somewhere on the case page that the tier 2 case is authorised for two way communication and can put stuff in the comments box.

jamesremuscat:
This PR allows contacts to comment on tier 1/2 cases, where they otherwise would not be able to, when permitted to do so by an admin.

Admins viewing open tier 1/2 cases will now see an option to enable commenting for contacts. For cases where commenting has been enabled, they will see a similar option to disable commenting.

Note that there's a display bug here: the `a.btn` within the `.alert` is being given a `text-decoration: underline;` that it shouldn't. I've fixed this already in one of the other branches I've been working on where this came up, so it'll be resolved when that branch gets merged into develop.

Closes #319.

Trello: https://trello.com/c/kcUUpdIh/341-allow-admins-to-enable-disable-user-communication-box-on-a-tier-1-2-case

### DIFF
--- a/app/models/case_commenting.rb
+++ b/app/models/case_commenting.rb
@@ -14,7 +14,7 @@ class CaseCommenting
       viewer_cannot_comment_message
     elsif !kase.open?
       not_open_message
-    elsif user.contact? && !kase.consultancy?
+    elsif user.contact? && !kase.consultancy? && !kase.comments_enabled
       non_consultancy_message
     else
       ''
--- a/app/models/case_commenting.rb
+++ b/app/models/case_commenting.rb
@@ -14,7 +14,7 @@ class CaseCommenting
       viewer_cannot_comment_message
     elsif !kase.open?
       not_open_message
-    elsif user.contact? && !kase.consultancy? && !kase.comments_enabled
+    elsif user.contact? && kase.comments_could_be_enabled? && !kase.comments_enabled
       non_consultancy_message
     else
       ''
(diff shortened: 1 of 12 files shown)


app/models/case_commenting.rb
```diff
@@ -14,7 +14,7 @@ def disabled_text
       viewer_cannot_comment_message
     elsif !kase.open?
       not_open_message
-    elsif user.contact? && !kase.consultancy?
+    elsif user.contact? && !kase.consultancy? && !kase.comments_enabled
```
bobwhitelock: Might be worth covering this new condition in `spec/models/case_commenting_spec.rb`.

app/views/cases/_case_comment_toggle_controls.html.erb
```diff
@@ -0,0 +1,15 @@
+<% if current_user.admin? && @case.open? && !@case.consultancy? %>
```
bobwhitelock: This conditional seems somewhat similar/related to the logic in `CaseCommenting#disabled_text` - we essentially want to display the controls here iff `current_user.admin? && contact_can_sometimes_comment?`. It would be nice if we could extract some common condition between these to avoid the possibility these controls being/not being displayed getting out-of-sync with when the `comments_enabled` flag will be considered when displaying the page to contacts - however how this should be done is not immediately obvious to me.

Possibly something to think about, though if nothing seems obvious to you either then I imagine this will be fine; we always only show these controls to admins anyway so if we later start showing them too often/not enough this should become obvious to us, and shouldn't effect non-admins in any case.

jamesremuscat:
Addressed both your comments, and rebased on `develop` to resolve conflict.

bobwhitelock:
Looks great, thanks :slightly_smiling_face:

PR #348, opened 2018-06-12
Handle invalid SSO tokens without raising HTTP 500 errors

jamesremuscat:
This PR handles the case where an invalid (either syntactically or semantically) SSO token was leading to a server error, by instead making `User.from_jwt_token` return `nil` in those cases.

I've also added a couple of tests for that method.

Trello: https://trello.com/c/xdF4dcDH/349-fix-500-when-sso-token-verification-fails


bobwhitelock:
LGTM, thanks

PR #351, opened 2018-06-13
Maintenance nearing completion notifications

atoghill:
This PR implements notifications via email and Slack for maintenance that is ending within an hour. Whenever maintenance windows are checked for progression it will also check if maintenance is ending in the next hour as well as if notifications have already been sent for this.

Notifications will only be sent for maintenance that satisfies the following conditions:
* Maintenance has started
* Maintenance is ending in the next hour
* Notifications have not been sent already

However if the maintenance window is extended then the flag for the notifications is reset allowing them to be sent the next time maintenance is near completion.

[Trello](https://trello.com/c/U0W6FE23/316-maintenance-is-nearing-completion-notification)


app/services/progress_maintenance_window.rb
```diff
@@ -65,4 +67,19 @@ def end_date
   def format_datetime(datetime)
     datetime.to_formatted_s(:short)
   end
+
+  def check_remaining_time
+    unless window.maintenance_ending_soon_email_sent
+      return unless 1.hour.from_now >= window.expected_end && started?
```
bobwhitelock: This method does slightly more than checking the remaining time so could be better named something like `maybe_send_maintenance_ending_soon_email`. Also do we need the nested `unless`es? Could make things a bit simpler if we collapsed them into one, and we could then pull out the whole condition into a method which we use as an early return. E.g. combining these would give something like:

```ruby

  def maybe_send_maintenance_ending_soon_email
    if should_send_maintenance_ending_soon_email?
      send_maintenance_ending_soon_email
    end
  end

  def should_send_maintenance_ending_soon_email?
    return false if window.maintenance_ending_soon_email_sent
    1.hour.from_now >= window.expected_end && started?
  end

  def send_maintenance_ending_soon_email
    CaseMailer.maintenance_ending_soon(
      window.case,
      <<-EOF.squish
        Maintenance for #{window.associated_model.name} is scheduled to end at
        #{window.expected_end.to_formatted_s(:short)}. You have less than an hour
        to make any final changes
      EOF
    )
    window.update!(maintenance_ending_soon_email_sent: true)
  end
```

app/services/progress_maintenance_window.rb
```diff
@@ -65,4 +67,19 @@ def end_date
...
+      CaseMailer.maintenance_ending_soon(
+        window.case,
+        <<-EOF.squish
+          Maintenance for #{window.associated_model.name} is scheduled to
+          end at #{window.expected_end.to_formatted_s(:short)}. You have
+          less than an hour to make any final changes
+        EOF
```
bobwhitelock: If this is the standard text we want for a 'maintenance ending soon' email then it might be better living in the `CaseMailer` method or possibly even better the email template (since it seems like a view concern)? I'd imagine if we do want to send the email in another context then we'd want to use the same or very similar text.

spec/mailers/case_mailer_spec.rb
```diff
@@ -165,7 +165,8 @@
     let (:text) { "Doesn't look like anything to me" }
 
     it 'sends a notification to Slack' do
-      expect(SlackNotifier).to receive(:maintenance_notification).with(kase, text)
+      expect(SlackNotifier).to receive(:maintenance_notification)
+        .with(kase, text, false)
```
bobwhitelock: Would be useful to have test coverage for `CaseMailer#maintenance_ending_soon` too - should be fairly straightforward to do and will look similar to the other tests in this file, with assertions that the text looks correct, it's only sent to admins etc.

app/services/progress_maintenance_window.rb
bobwhitelock: Also: I'd end the second sentence with `.` :stuck_out_tongue:

spec/services/progress_maintenance_window_spec.rb
```diff
@@ -119,5 +119,44 @@ def test_progression_message(message:, window:, expected:)
...
+        it 'sends notifications that maintenance is ending soon' do
+          window = build_window(state: :started)
+
+          expect(CaseMailer).to receive(:maintenance_ending_soon)
+
+          described_class.new(window).progress
+        end
+      end
+
+      context 'when maintenance_ending_soon_email_sent is true' do
```
bobwhitelock: This context is also checking things in the 'maintenance has started' case, and contains the same `window = build_window(state: :started)` line, so you could nest it within the above context and then extract a shared `let :window` used by both examples.

spec/models/maintenance_window_spec.rb
```diff
@@ -230,6 +230,15 @@
 
         subject.extend_duration!(create(:admin))
       end
+
+      it 'resets the maintence nearing completion email sent flag' do
```
bobwhitelock: `s/maintence/maintenance/`

spec/services/progress_maintenance_window_spec.rb
```diff
@@ -119,5 +119,44 @@ def test_progression_message(message:, window:, expected:)
       other_states = MaintenanceWindow.possible_states - [:started, :confirmed, :new, :requested]
       include_examples 'does not progress', other_states
     end
+
+    describe 'maintenance is nearing completion' do
```
bobwhitelock: We're not checking the case where the maintenance is `started`, the email hasn't already been sent, but we're not in the last hour of the maintenance here - I think this should be fairly straightforward to add, the assertion should be the same as the `when maintenance_ending_soon_email_sent is true` case just with slightly different setup.

app/services/progress_maintenance_window.rb
```diff
@@ -65,4 +67,19 @@ def end_date
...
+      )
+      window.update!(maintenance_ending_soon_email_sent: true)
```
bobwhitelock: We're not currently checking this flag is set in the tests, which I think is quite an important part of this functionality since this is what stops us being sent the email every minute for the hour until the maintenance ends, and so would be worth adding. The minimal way to add coverage for this would be to add an assertion to the 'email sent' test case that this flag is `true` afterwards. However, for a better separation of concerns I'd suggest the following:

1. add a method, say `set_maintenance_ending_soon_email_flag`, to `MaintenanceWindow`, containing this line;

2. call this method from `CaseMailer#maintenance_ending_soon` and test that this sets the flag in its own tests;

3. remove this line from this class, and the tests for this class shouldn't need changing.

This would have the following advantages imo:

1. it confines knowledge of how the `maintenance_windows` table is structured, in particular that it has a boolean `maintenance_ending_soon_email_sent` field, to the `MaintenanceWindow`s class; this is generally a good practise as it means the knowledge of the database is spread in less places and so there's less places to update if this changes;

2. it directly couples sending the 'maintenance ending soon' email with setting the flag that this has been sent, which is useful as wherever we may want to send this from this will ensure that the flag is set; which I'd imagine is the behaviour we want and should avoid the possibility of double-sending.

However I recognise that this would be quite a bit more work, and this is slightly academic atm as we're only sending the email here; therefore, as with other comments, if you'd prefer to go for the quicker fix here that's fine by me.

app/services/progress_maintenance_window.rb
```diff
@@ -2,8 +2,10 @@
 ProgressMaintenanceWindow = Struct.new(:window) do
   def progress
     if required_transition_event
+      check_remaining_time
       transition_state(required_transition_event)
     else
+      check_remaining_time
```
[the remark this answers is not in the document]
atoghill: Didn't used to be able to as this would break a bunch of tests however it should (?) be fine now that I have made changes to where that was breaking. Will investigate and change if necessary

app/services/progress_maintenance_window.rb
atoghill: @bobwhitelock I've made changes to everything but the first comment here. I wanted things to fit to the current format of methods within the `CaseMailer` so I was avoiding putting the text directly within those. However it could be worth just throwing the text within the email itself, instead of passing it to it like `maintenance_state_transition` does.

bobwhitelock:
Looks great, thanks @atoghill, just made a couple of extra tweaks and merging now :shipit:

PR #352, opened 2018-06-13
Clickable case row

atoghill:
This PR highlights and makes the case row clickable. On click the user is redirected to the associated case page, just like the link via the case ID on the current site. However the links in the `Association` column still function as before.

During review it would be worth experimenting with the opacity within the `case-highlight` class found within `cases.scss`. You may find it looks better without any change in opacity but the option is there if you did want it that way.

It is also worth checking that links in the row are still distinct enough to stand out. I experimented with various css changes to try and find something that worked but it may not be enough.

[Trello](https://trello.com/c/tzcE8SSw/350-make-clicking-anywhere-on-case-row-in-table-link-to-case-page)

### DIFF
--- a/app/views/partials/_case_table_row.html.erb
+++ b/app/views/partials/_case_table_row.html.erb
@@ -15,6 +15,8 @@
         <%= kase.assignee&.name || 'Nobody' %>
       <% end %>
     </td>
+    <%# The below onclick makes sure that the onlick event within the row does
+      not get triggered when clicking the link within this column. %>
   <td onclick='event.stopPropagation();'><%= kase.association_info %></td>
   <% if kase.resolved? || kase.closed? %>
     <td><%= kase.credit_charge&.amount %></td>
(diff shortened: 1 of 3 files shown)


app/views/partials/_case_table_row.html.erb
```diff
@@ -1,20 +1,21 @@
...
+    <td style="white-space: nowrap;"><%= kase.user.name %></td>
+    <td><%= kase.subject %></td>
+    <td>
+      <% if current_user == kase.assignee %>
+        <%= raw('<strong>Me</strong>') %>
+      <% else %>
+        <%= kase.assignee&.name || 'Nobody' %>
+      <% end %>
+    </td>
+  <td onclick='event.stopPropagation();'><%= kase.association_info %></td>
```
bobwhitelock: Possibly worth adding a comment about why we want this `onclick` here (so that the links within this column can be followed without triggering the `tr` `onclick`), as I can imagine this not being obvious to someone who wasn't there for our original discussion.

atoghill: Yep that is a very good point, I'll add a comment to this 👍

PR #353, opened 2018-06-14
Markdown previewing for notes

benarmston:
This PR adds support for previewing the rendered markdown for cluster notes.  It does so by submitting the form to the usual action `create` or `update`.  The action branches based on which button was used to submit the form, the "preview" button, the "write" button or the "submit" button.  The page is then rendered again showing the preview.

This has been implemented in a fairly generic way, but support for case comments, logs and other models has not yet been added.  It may be necessary to change to the mechanism used to generate the preview for them as they have many other things on their respective pages.


benarmston:
Probably best to ignore this PR in favour of #362 .

PR #354, opened 2018-06-14
Use a styled `div` rather than `blockquote` in emails.

### LINKED ISSUE #335: blockquote tag in email confuses Airmail email client
I use [Airmail](http://airmailapp.com/) which is a popular email application on MacOS.

The Flight Center emails don't always display correctly (see attached).  This seems mainly to do with the fact we're using `blockquote` which it's interpreting as an instruction to fold up the content as if it were from a previous email.

The content can be unfolded, though frequently this causes the email rendering to break entirely and display as blank.

Could we try changing the template to use a styled `div` or `p` instead?

(See [#alces-flight-center for some related screenshots](https://alces.slack.com/archives/C72GT476Y/p1528274758000450).)

jamesremuscat:
Hopefully this will stop Mark's email client from getting confused.

Fixes #335.

Trello: https://trello.com/c/VoWTgepK/348-blockquote-tag-in-email-confuses-airmail-email-client

### DIFF
--- a/app/views/layouts/mailer/_inline_main.scss
+++ b/app/views/layouts/mailer/_inline_main.scss
@@ -118,7 +118,7 @@ code {
     padding: 2px 4px;
 }
 
-blockquote {
+blockquote, div.blockquote {
     padding: 1rem;
     border-radius: 5px;
     box-shadow: 0 0 5px $c-primary-color;
--- a/app/views/layouts/mailer/_inline_main.scss
+++ b/app/views/layouts/mailer/_inline_main.scss
@@ -118,7 +118,7 @@ code {
     padding: 2px 4px;
 }
 
-blockquote, div.blockquote {
+div.blockquote {
     padding: 1rem;
     border-radius: 5px;
     box-shadow: 0 0 5px $c-primary-color;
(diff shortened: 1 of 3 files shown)


app/views/layouts/mailer/_inline_main.scss
```diff
@@ -118,7 +118,7 @@ code {
     padding: 2px 4px;
 }
 
-blockquote {
+blockquote, div.blockquote {
```
bobwhitelock: Do we still want to match `blockquote` tags (quite possible that we do, you know more than me about our email styling :slightly_smiling_face:)? If not could be worth removing this, I can imagine a situation where we re-introduce the same issue for a new email by using the tag rather than the class, which we might not notice if we're still styling the tag in the same way as the class.

mjtko:
FYI: it will need to be two emails for me to reproduce/verify the fix as it happens when rendering the "conversation".

app/views/layouts/mailer/_inline_main.scss
jamesremuscat: We don't use it anywhere else in FC so might as well get rid of that definition, as you say.

bobwhitelock:
Thanks for making that change @jamesremuscat :slightly_smiling_face:.

@mjtko we've decided not to spam you from development and I'm going to merge this now which will close the corresponding issue, please re-open it if you still encounter it once we've deployed to staging/production.

PR #355, opened 2018-06-14
Use custom error pages and improve need-to-log-in experience

jamesremuscat:
This PR supersedes #95 to improve the user experience of being directed to a page requiring a log-in.

We now show a properly themed 404 page, complete with Flight chrome, which suggests that the user try logging in (if they have not already done so). Logged in users are not shown this message.

Note that we currently don't differentiate between "user is logged in to Flight but has no Flight Center account" and "user is not logged in" - that's probably worth doing, here and on the home page, but doesn't really fit into a PR about custom error pages.

Trello: https://trello.com/c/l5msCYDI/116-if-try-to-visit-page-while-logged-out-redirect-to-page-after-login


bobwhitelock:
This is great, thanks @jamesremuscat. Just made one addition to document how to enable these pages, and will now merge :slightly_smiling_face:.

PR #357, opened 2018-06-15
Distinguish case history cards more

atoghill:
The changes made in this PR are somewhat subtle but that is by design. Making the cards more distinguishable without making them intrusive and ugly proved problematic. Therefore I opted with using Boostrap Tooltips that give more context to the icons when the user hovers over them. Hopefully this should clear any confusion that comes from these.

This had the knock on effect of making other areas where we use `title` as a tooltip to use this Bootstrap styling for them. Which should make those look a bit nicer than before as well.

[Trello](https://trello.com/c/Fvlb0KFq/352-consider-distinguishing-case-history-cards-more)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -18,8 +18,8 @@
 //= require_tree .
 //
 
-function enable_tooltips() {
+function enableTooltips() {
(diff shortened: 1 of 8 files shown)


bobwhitelock:
This looks great to me @atoghill, and nice that we can get the nicer looking tooltips everywhere we were showing `title` text before. :shipit:

PR #360, opened 2018-06-18
Remove unwanted code

benarmston:
In the tool preselection PR #346 it was decided that this code should be removed as there are a couple of edge cases and currently no requirement for the functionality.  It is now belatedly removed.


bobwhitelock:
Thanks @benarmston!

PR #362, opened 2018-06-18
Markdown preview all the things

benarmston:
This PR adds markdown preview functionality for logs and case comments and changes the implementation for previewing notes.  The implementation for notes previewing has changed to be more generic / reusable.

Viewing the markdown preview is accomplished by clicking on the "Preview" tab.  This click is intercepted by a small piece of javascript, which makes an XHR POST request to the appropriate URL.  The response is the markdown preview and the wrapping "Write" / "Preview" tabs.  The javascript then replaces the text area and wrapping "Write" / "Preview" tabs with the response.

Going back to writing the markdown is accomplished by clicking on the "Write" tab.  This is handled similarly to clicking on the "Preview" tab.


jamesremuscat:
@benarmston / @bobwhitelock - any chance that one of you could rebase this on `develop`, so that it's done by someone familiar with at least half of the changes involved? :slightly_smiling_face:

bobwhitelock:
@jamesremuscat I think we've both independently tried now and run into issues with lack of knowledge of the changes the other one made :stuck_out_tongue:. Probably makes most sense to do it together so we can discuss as we go, which would be easiest next time we're both in the office - which should be next week (unless of course you need to make changes to markdown/note things sooner than that, in which case I can have another attempt sooner).

benarmston:
@jamesremuscat rebasing was certainly a real pain.  It may require @bobwhitelock and I to do it together.

benarmston:
@jamesremuscat @bobwhitelock I've now rebased this onto develop.

I think that all of the permission checks are correct, but it would be a good idea to give those a bit more scrutiny.

bobwhitelock:
Thanks for doing that @benarmston, from a quick check the permissions checking for notes looks like it's still working as expected :slightly_smiling_face:.

jamesremuscat:
I made a small change so that the edit form uses the policy for `update` rather than `create` (not that those are any different). If you're both happy then I'll merge this through now :slightly_smiling_face:

bobwhitelock:
@jamesremuscat Possibly worth tweaking/removing the comment at https://github.com/alces-software/alces-flight-center/pull/362/commits/b6c1ea4a6fd58f9beefc5dabe6ed078bf75ebcd2#diff-ef0916f6b5e43464a33f1c9a21ed01afR20 following that change since the 'create-specific stuff' has now been removed :slightly_smiling_face:. And happy for you to merge this when you're happy with it :shipit:.

jamesremuscat:
@bobwhitelock Ha, my IDE had "helpfully" collapsed that comment! I think it's still relevant, so I've updated its wording.

PR #363, opened 2018-06-18
Cluster shortcode in links

atoghill:
This PR utilises the cluster shortcode to create a vanity URL for cluster related links.

[Trello](https://trello.com/c/TuSECh2A/357-use-cluster-shortcode-in-links-to-cluster)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -72,7 +72,7 @@ class ApplicationController < RootController
     @scope = case request.path
              when /^\/clusters/
                id = scope_id_param(:cluster_id).upcase
-               @cluster = Cluster.find_by_shortcode(id)
+               @cluster = Cluster.find_by_shortcode!(id)
              when /^\/components/
                id = scope_id_param(:component_id)
                @component = @cluster_part = Component.find(id)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -72,7 +72,11 @@ class ApplicationController < RootController
     @scope = case request.path
              when /^\/clusters/
                id = scope_id_param(:cluster_id).upcase
-               @cluster = Cluster.find_by_shortcode!(id)
+               @cluster = if /^[0-9]+$/.match(id)
+                            Cluster.find(id)
+                          else
+                            Cluster.find_by_shortcode!(id)
+                          end
              when /^\/components/
                id = scope_id_param(:component_id)
                @component = @cluster_part = Component.find(id)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -72,11 +72,7 @@ class ApplicationController < RootController
     @scope = case request.path
              when /^\/clusters/
                id = scope_id_param(:cluster_id).upcase
-               @cluster = if /^[0-9]+$/.match(id)
-                            Cluster.find(id)
-                          else
-                            Cluster.find_by_shortcode!(id)
-                          end
+               @cluster = Cluster.find_from_id!(id)
              when /^\/components/
                id = scope_id_param(:component_id)
                @component = @cluster_part = Component.find(id)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -71,7 +71,7 @@ class ApplicationController < RootController
 
     @scope = case request.path
              when /^\/clusters/
-               id = scope_id_param(:cluster_id)
+               id = scope_id_param(:cluster_id).upcase
                @cluster = Cluster.find_by_shortcode(id)
              when /^\/components/
                id = scope_id_param(:component_id)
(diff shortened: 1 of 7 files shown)


jamesremuscat:
Just a note that we'll probably want the URLs to be case-insensitive, so that `/clusters/ham`, `/clusters/HAM` and `/clusters/hAm` all go to the same place.

atoghill:
@jamesremuscat Yep I agree. Is it acceptable to simply have those example URL's go the same page or do you think the URL should change to `/clusters/HAM` in all cases?

app/controllers/application_controller.rb
```diff
@@ -72,7 +72,7 @@ def assign_scope
     @scope = case request.path
              when /^\/clusters/
                id = scope_id_param(:cluster_id)
-               @cluster = Cluster.find(id)
+               @cluster = Cluster.find_by_shortcode(id)
```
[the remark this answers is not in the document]
bobwhitelock: I note you've now already done what I suggested, so you can ignore this comment :slightly_smiling_face:.

app/controllers/application_controller.rb
```diff
@@ -71,8 +71,8 @@ def assign_scope
 
     @scope = case request.path
              when /^\/clusters/
-               id = scope_id_param(:cluster_id)
-               @cluster = Cluster.find(id)
+               id = scope_id_param(:cluster_id).upcase
+               @cluster = Cluster.find_by_shortcode(id)
```
bobwhitelock: I think we should use the bang method, `find_by_shortcode!` here, so that when the cluster isn't found we get a 404 rather than a `NoMethodError` in a distant place due to the cluster being `nil` (this will be the same behaviour as we had before).

This is another one of those confusing/inconsistent Rails things - the regular `find` method will raise an error which will give a 404 if the record isn't found, but the `find_by_*` versions will just give `nil` if the record isn't found, and you need to use `find_by_*!` to get the same behaviour.

spec/requests/navigation_variable_assignments_spec.rb
```diff
@@ -34,7 +34,7 @@ def assigns_navigation_variables(*subject_keys)
     describe "get '/cluster/*'" do
       subject { cluster }
       before :each do
-        get cluster_path(cluster.id, as: user)
+        get cluster_path(cluster.shortcode, as: user)
```
bobwhitelock: In this and various other places where we're using Cluster path helpers we've had to update our usage to pass the Cluster `shortcode` rather than `id` - however if in each case we instead just pass the Cluster instance itself and let the path helper appropriately handle this by `to_param`-ing it, I don't think we'll need to do this. This will both make these method calls simpler and means we won't need to change these method calls again if we change the format for Cluster URLs again. (pointing this out both as a suggestion for a change you could make now and as something to consider when using path helpers in future :slightly_smiling_face:).

bobwhitelock:
Also: there was a failing test before - not sure if this was a real issue or just JS tests being flaky, so I've restarted CI to be sure. If it fails again will need looking into.

app/models/cluster.rb
```diff
@@ -88,6 +88,10 @@ def credit_balance
     end
   end
 
+  def to_param
+    shortcode
```
jamesremuscat: We should be calling `parameterize` on `shortcode` before returning it here (in the unlikely event of us using special characters in a shortcode, URLs would break otherwise).

atoghill: Neat thanks for informing me of that method

atoghill:
Rebased on `develop` and fixed merge conflicts

atoghill:
Rebased on `develop`

jamesremuscat:
Sorry to not have picked up on this before, but this change will invalidate all current URLs involving clusters with numeric IDs.

We should do as we do for Cases; that is, try finding by ID if the parameter is numeric, else assume it's a shortcode and find by that.

See https://github.com/alces-software/alces-flight-center/blob/develop/app/models/case.rb#L114 (which, on review, doesn't need to sometimes call `.decorate` and sometimes not...)

atoghill:
Is that not the point of this change? The way I saw things was that we wanted to stop using the `cluster_id` for URLs and instead use the `shortcode`. This was not only to make the URLs more meaningful but also to obfuscate the number of clusters we managed.

I may be reading your comment completely wrong so please correct me if I'm wrong/mistaken here.

jamesremuscat:
@atoghill We do want to stop using them, but if customers have bookmarks/emails/other references using the old URLs, we don't want to start sending them to a 404. While _we_ won't be _generating_ numeric-ID URLs any more, we can't guarantee that _someone_ won't try and use one at some point (though that likelihood will diminish over time).

In general, it's good to ensure backwards-compatibility where possible (and, as I say, the change we made to Cases is a precedent).

atoghill:
@jamesremuscat Gotcha! That makes more sense now, cheers. I'll push a commit for this soon.

Issue #207, opened 2018-04-29
Flight Center version display


ste78:
I guess flight center has a version or tag or something? we should consider putting this in the footer so we know which version we are testing.

bobwhitelock:
Trello: https://trello.com/c/u07sGH6U/174-flight-center-version-display

bobwhitelock:
Brief idea for how to do this: in `alces:deploy:*` Rake tasks, after successfully deploying app set an environment variable in Dokku with same tag as pushed to GitHub; this can then be grabbed by the app and displayed in the footer (or wherever).

PR #370, opened 2018-06-20
Version display

atoghill:
This PR adds a version display just above the footer. Initially I have this in the format of `<year>.<version_number> <version name>` where `version_name` is a two word alliteration based on the `version_number` (If this equals 4 then use the 4th letter). With the first word being a verb and the second a bird.

As it stands the `version_name` shows up everywhere but `Production`. However this can easily be changed, as well as the actually formatting of the version display.

Note: There is currently nowhere stating that this needs to be changed between releases so finding a suitable place for that would be great

[Trello](https://trello.com/c/u07sGH6U/318-flight-center-version-display)

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -128,6 +128,14 @@ table {
 }
 
 .version {
-  font-size: 10px;
-  text-align: center;
+  font-size: 13px;
+  color: white;
+  margin-bottom: 0; // Remove default margin from p element
+
+  // This should push the text to the bottom right of the page within the
+  // footer
+  position: relative;
+  float: right;
+  top: -2.2em;
+  right: 1em;
 }
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -126,3 +126,8 @@ table {
 .alert a.btn {
   text-decoration: none;
 }
+
+.version {
+  font-size: 10px;
+  text-align: center;
+}
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -50,8 +50,8 @@
 
     </div>
 
-    <p class="version"> 2018.4 <%= 'Daydreaming Dickcissel' unless Rails.env.production? %></p>
     <div id='brand-footer'></div>
+    <p class="version"><%= Rails.application.config.version %></p>
     <script>
       window.__SSO_BASE_URL__ = '<%= Rails.application.config.sso_base_url %>';
       window.__REFRESH_AFTER_LOGIN__ = true;
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -50,6 +50,7 @@
 
     </div>
 
+    <p class="version"> 2018.4 <%= 'Daydreaming Dickcissel' unless Rails.env.production? %></p>
     <div id='brand-footer'></div>
     <script>
       window.__SSO_BASE_URL__ = '<%= Rails.application.config.sso_base_url %>';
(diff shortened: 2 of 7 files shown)


bobwhitelock:
Couple of quick things:
- I think the original reasoning behind displaying a version number is because Steve wanted to see what was running in production, so would make sense to display this there;

- currently this is a hard-coded string with no relation to the actual version number tagged (as in https://github.com/alces-software/alces-flight-center/releases), would make sense to unify these to display the same string otherwise this version displayed in the app is only likely to be misleading.

atoghill:
Addressing your first point @bobwhitelock there is a version displayed in every environment. It is just the `version_name` that I have omitted from `Production`

atoghill:
Rebased on `develop` and fixed merge conflict

app/views/layouts/application.html.erb
```diff
@@ -49,6 +49,7 @@
 
     </div>
 
+    <p class="version"> 2018.4 <%= 'Daydreaming Dickcissel' unless Rails.env.production? %></p>
```
jamesremuscat: At the very least, this should be a configuration setting defined in `config/application.rb` rather than a hardcoded string somewhere in the middle of our layout template.

jamesremuscat: Let's ditch the alliterative name concept, that's just going to be a pointless time sink.

app/assets/stylesheets/application.scss
```diff
@@ -126,3 +126,8 @@ table {
 .alert a.btn {
   text-decoration: none;
 }
+
+.version {
```
jamesremuscat: I'm not particularly happy with the way this looks right now - it looks out of place where it is. I did suggest something fixed to the bottom-right of the viewport (over the top of the Flight footer when the page is short), see attached for an example of what I meant.

alces flight center - google chrome_028

app/views/layouts/application.html.erb
atoghill: No fun

app/assets/stylesheets/application.scss
atoghill: Yep you did and I did have an implementation similar to that but felt that it looked even more out of place there. Where it is currently is unobtrusive in my opinion but still easy enough to find if you really wanted to know what version was running. 

You're more than welcome to make the executive decision here but I implemented what I felt was the best way of handling this (Without going in to the React stuff that I was advised to stay away from)

ste78: can somebody attach a screenshot of how it looks atm?

atoghill: screenshot-2018-6-21 alces flight center

ste78: Had to google "dickcissel" but yeah sadly we should drop that :(

Without the black background looks fine imo but I agree that fully over to the right would make it more subtle.  

Pinned to the footer as James suggested would probably be preferable but I'm guessing that looks weird when the page content is the full screen width?? We don't really want it to overlay actual content, so if all the way down the bottom is the best we can do to avoid that it'll have to do :/

atoghill:
Rebased on `develop`

@jamesremuscat The latest commits should hopefully satisfy the requirements you wanted a bit better now. I managed to place the version name on the bottom right of the footer like so:

version-display-footer

Also this should now be handled via the `rake` deploy task therefore when running that command you would now need to specify the version name using `.. VERSION='blah'`. I have tested this using a `dry_run` but let me know if I have broken anything.

jamesremuscat:
Looks great now, thanks. I've made one small further change - I've moved the call to `config:set` earlier in the process and added the `--no-restart` flag (that you wouldn't have known about). This means that the version number change is picked up when the app is redeployed immediately thereafter, and we don't trigger a subsequent restart afterwards.

Will merge when CI has completed.

atoghill:
Excellent, I wasn't sure where the best point in the process to put it was so I'm glad you moved it. But you're right I did not know about the restart flag so TIL 👍 

Note: Once merged this should resolve #207

PR #376, opened 2018-06-21
Prevent missing event details from breaking case page

atoghill:
When I added tooltips to icons for case events in #357 I didn't account for new event cards missing a `details` key-value pair. However this should now assign a nil value to `details` if it is unable to find a value for the event card and therefore would not display any tooltip when hovering over the affected icon.

### DIFF
--- a/app/views/cases/_event.html.erb
+++ b/app/views/cases/_event.html.erb
@@ -2,7 +2,7 @@
   <div class="card event-card">
     <div class="card-header d-inline-flex justify-content-between align-items-center">
       <span><%= date.to_formatted_s(:long) %> - <b><%= name %></b></span>
-      <div title="<%= details %>">
+      <div title="<%= details ||= nil %>">
           <%= icon(type) %>
       </div>
     </div>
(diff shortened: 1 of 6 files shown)


app/views/cases/_event.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="card event-card">
     <div class="card-header d-inline-flex justify-content-between align-items-center">
       <span><%= date.to_formatted_s(:long) %> - <b><%= name %></b></span>
-      <div title="<%= details %>">
+      <div title="<%= details ||= nil %>">
```
jamesremuscat: In general, we've had a policy of "fail fast" in these situations, so that when we add new event cards we start failing tests straight away when there's information missing (`details` in this case).

That this hasn't happened implies that there's no test that tries to create an event card for change requests - so please could you add such a test and revert this line?

atoghill: That wouldn't stop future event cards from causing issues. The point here is that when a new event card has been made without `details` it would break and unless people know to make a test for it then it could slip through like this did.

atoghill:
@jamesremuscat 9b28bda added a test and reverted that change like you asked for. I still think problems could arise from event cards in the future as this system is wholly reliant on a test being made in the first place.

Also rebased on `develop`

jamesremuscat:
@atoghill I've added a test in 4783c59 that checks that, for each `Decorator` that has an `event_card` method, that method can be called without errors being raised. Hopefully this will assuage your concerns.

atoghill:
Yep that looks great @jamesremuscat, cheers for adding that :)

PR #377, opened 2018-06-22
Change request event Slack notifications

atoghill:
This PR adds Slack notifications for whenever a change request event sends an email out. These will be displayed like so:

change-request-slack

With links to both the associated case and the change request itself.

[Trello](https://trello.com/c/8XlyhP8U/359-add-slack-notifications-to-change-request-events)

### DIFF
--- a/app/mailer_previews/case_mailer_preview.rb
+++ b/app/mailer_previews/case_mailer_preview.rb
@@ -23,7 +23,8 @@ class CaseMailerPreview < ApplicationMailerPreview
   def change_request
     CaseMailer.change_request(
       get_case,
-      'has text to be replaced with the text from ChangeRequestStateTransitionDecorator.'
+      'has text to be replaced with the text from ChangeRequestStateTransitionDecorator.',
+      'User'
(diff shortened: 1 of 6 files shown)


atoghill:
Rebased on `develop`

app/mailer_previews/case_mailer_preview.rb
```diff
@@ -23,7 +23,8 @@ def maintenance
   def change_request
     CaseMailer.change_request(
       get_case,
-      'has text to be replaced with the text from ChangeRequestStateTransitionDecorator.'
+      'has text to be replaced with the text from ChangeRequestStateTransitionDecorator.',
+      'User'
```
jamesremuscat: This should be an instance of `User` rather than a string - as it is, the email preview throws an error. (It's probably acceptable to assume that there's at least one `User` in the database.)

atoghill: Yup this was something I was unsure about, should as of d05413e.

PR #385, opened 2018-06-28
Significantly improve Case form app compile time

bobwhitelock:
By default Rails' Webpacker will build everything in
`app/javascript/packs`, and rebuild everything when any file in this
directory changes. It is intended that this directory should just be
used for the entry points of apps to be compiled by Webpack, with the
majority of actual app code living in `app/javascript`.

However at the point I ran the Elm Webpacker installer it was generating
an `elm-package.json` which only looks in `app/javascript/packs`, and so
I continued adding new files directly to this directory.

While developing https://github.com/alces-software/flight-inventory I
noticed I was getting significantly faster compile times, due to placing
files in `app/javascript` instead, which is where the
Webpacker-generated `elm-package.json` now looks by default. By changing
our Case form app to match this, with all Elm files in `app/javascript`,
I can also get significantly faster compile times for this. I believe
this is something related to Webpack now not needing to independently
compile every Elm file we have, and instead it can just compile `Main`,
and from this everything directly or indirectly imported by `Main`.


jamesremuscat:
Nice. My CPU is now considering how best to spend all those cycles you just freed up :slightly_smiling_face:

PR #448, opened 2018-07-24
Show open cases only at "Manage site cases" / "View cases" links.

### LINKED ISSUE #447: View cases/Manage support cases links should link to open only
There's been a change in behaviour since the filtering was introduced: clicking on the "Manage support cases" link (for admins) or the "View cases" button from the site dashboard now shows all cases by default, including resolved/closed cases.  Previously this only showed open cases.

Can we update these buttons/links to return to the previous behaviour please (especially as loading/rendering all cases is quite slow at the moment, ref #446).

jamesremuscat:
Fixes #447.


jamesremuscat:
Cherry-picked to `release/2018.5`.

PR #450, opened 2018-07-24
Don't initialise `collapse`s on page transition.

### LINKED ISSUE #449: Cosmetic: buttons seem to animate into existence when rendered
I've noticed that the buttons rendered in the brand bar ("sign in") and the nav bar ("all sites", site name, cluster name etc.) have an odd "vertically expand from middle animation" on page transition.

This doesn't appear to be the case on the current live system -- perhaps an odd animation style has leaked out to places it should not during the last iteration?

jamesremuscat:
We don't need to initialise the Bootstrap `Collapse`s used by the cluster tree, since they are initialised individually by the `_init_cluster_tree` function. Doing so was interacting badly with the Flight chrome navbar which also uses a `.collapse`.

Fixes #449.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -26,10 +26,5 @@ function enablePopovers() {
   $('[data-toggle="popover"]').popover();
 }
 
-function enableCollapses() {
-  $('.collapse').collapse();
-}
-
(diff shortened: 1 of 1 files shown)


jamesremuscat:
Has been cherry-picked into `release/2018.5`.

PR #460, opened 2018-07-30
Remove unneeded MOTD presence validation

bobwhitelock:
This was added in
https://github.com/alces-software/alces-flight-center/commit/1c192659,
however prior to being added several Clusters were created with empty
MOTDs. This was then causing these Clusters to be invalid, which in turn
was causing submitting the new Case form for these Clusters to fail.

Since:

a. things have otherwise been working fine with blank Cluster MOTDs
until this validation was added;

b. from a business perspective, Clusters are allowed to not have MOTDs,
since these Clusters were created without a MOTD set,

we have decided to remove this validation.

Also removed this field from the Cluster factory, since this is no
longer required to create a minimal Cluster, which then prompted the
need to handle the case where the MOTD is null in the new Case form (by
these tests failing).

More discussion: https://alces.slack.com/archives/C72GT476Y/p1532949789000171.

Trello: https://trello.com/c/c6CgMcok/412-remove-unneeded-motd-presence-validation.

### DIFF
--- a/app/javascript/Cluster.elm
+++ b/app/javascript/Cluster.elm
@@ -47,13 +47,16 @@ type Id
 
 decoder : D.Decoder Cluster
 decoder =
-    D.field "motd" D.string
+    D.field "motd" (D.nullable D.string)
         |> D.andThen decodeWithMotd
 
 
-decodeWithMotd : String -> D.Decoder Cluster
-decodeWithMotd motd =
+decodeWithMotd : Maybe String -> D.Decoder Cluster
+decodeWithMotd maybeMotd =
     let
+        motd =
+            Maybe.withDefault "" maybeMotd
+
         serviceDecoder =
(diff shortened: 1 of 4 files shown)


bobwhitelock:
This will also need backporting to staging once you're happy with merging this.

bobwhitelock:
@jamesremuscat Want me to backport now?

jamesremuscat:
I just did! Merged into `release/2018.5`.

bobwhitelock:
Ah lovely, thanks :slightly_smiling_face:. One thing to consider, likely not a big deal in this case: by manually cherry-picking into `release/2018.5` this branch is now (I believe) in a state which has never been run in CI, so could potentially have test failures due to conflicting backported changes. Could be worth checking this to be sure before release, if we're not already.

PR #463, opened 2018-07-30
change number of spent credits to positive

### LINKED ISSUE #390: Usage report shows '-1 credits' when a single credit is charged
There's a pluralization issue, but also I think this should simply say "1 credit" as it's obvious from the fact it's against a case that it's a charge, and it's in orange.

DavidMarchant:
Fixes #390.

Trello: https://trello.com/c/MaoR0t6n/408-credit-charges-in-credit-report-should-be-shown-as-positive-values

### DIFF
--- a/app/views/clusters/_credit_charge_entry.html.erb
+++ b/app/views/clusters/_credit_charge_entry.html.erb
@@ -1,4 +1,4 @@
-<li class="list-group-item credit-charge-entry <%= credit_value_class(amount) %>" <%=amount = amount.abs%>>
+<li class="list-group-item credit-charge-entry <%= credit_value_class(amount) %>">
   <span>
     <i class="fa fa-circle mr-2" aria-hidden="true"></i>
   </span>
@@ -9,6 +9,6 @@
     <%= yield %>
   </span>
   <span>
-    <%= pluralize(amount, 'credit') %>
+    <%= pluralize(amount.abs,'credit') %>
   </span>
 </li>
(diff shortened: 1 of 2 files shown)


app/views/clusters/_credit_charge_entry.html.erb
```diff
@@ -1,4 +1,4 @@
-<li class="list-group-item credit-charge-entry <%= credit_value_class(amount) %>">
+<li class="list-group-item credit-charge-entry <%= credit_value_class(amount) %>" <%=amount = amount.abs%>>
```
jamesremuscat: This is an _unorthodox_ way of doing it :wink: 

In Rails' `erb` templates, a `<%= ... %>` tag means 'print the result of this Ruby at this point in the HTML markup'. This addition means we're randomly adding numbers as attributes to the `li` tags! 

```html
<li class="list-group-item credit-charge-entry text-warning" 2>
```

A tag without the equals sign `<% ... %>` means 'run this Ruby and don't print the result'. 

In this case, however, it's sufficient just to print `amount.abs` at the point it's used on line 12 rather than modify one of the parameters passed to the template in this way.

DavidMarchant: Oh right, ok, very sorry, it was only at about 9pm yesterday I even found out that there were more than one kind of erb tag! I'll change that now & make sure i don't do something similar again

jamesremuscat:
LGTM, thanks!

For reference: you need to use a magic verb e.g. "Fixes" when mentioning issues, in order for Github's magic to work.

Issue #361, opened 2018-06-18
'Other' Service / category / issue for tier 3 comments


ste78:
We're having cases logged under strange categories in order to be able to get a large tier 2 entry box so that they can use it to explain a then unrelated issue. It's difficult to correct this unless we have a place for people to actually put something that they feel doesn't categorise into any of our pre-defined choices. 

I think we've discussed this before but decided to not do it, but I'd like to suggest having an 'Other' -> 'Other' -> 'Other' -> Tier 3 only case that we can ask people to use if they feel their issue really doesn't fit into our categories. Open to any other suggestions that may be easier to implement that still fix the same issue without also encouraging people to not attempt to categorise.

mjtko:
I would approach it something like this:

 * creating any new issue always has it as tier 3 with just an entry for the description
 * if a user makes an effort to make an association with a component or service (or, perhaps, multiple components and/or services), then tier 2 and tier 1 options pop up ("Wow! Such Free!")
 * the interface should somehow encourage users to make the effort :)

The potential problem is that everything starts coming through as tier 3 with just a description, so we'd need to mitigate that somehow, probably.

jamesremuscat:
At this point I think the following is uncontroversial:

- We should allow an "Other / N/A" selection for `Service`, which should correspond to null (e.g. no association with any service)
- We should create an "Other / N/A" `Issue` which only permits tier 3 cases and is always available

I think we can achieve this through data/server changes without needing to change the Elm case form (see ba57e481 for a rough first pass).

One outstanding decision to be made is whether these new "Other" options are the default selections (as Mark's comments suggest, thus by default with no effort customers can raise a tier 3 case) or if we want a "Please choose..." placeholder entry and not display options for any tier until customers actively make a selection (as per Bob's comment). Both would require making (I suspect minimal) changes to the Elm form.

An alternative, that would require larger changes to (or replacement of) the Elm form, would be to allow the same sort of multi-or-non-selection of services/components that we give to admins post-creation. This would also mean we need to rethink the whole way we structure cases, issues, and tiers, and come up with a new approach that makes sense. That's a far bigger change, and conceptual shift, than the relatively straightforward changes discussed above.

mjtko:
Depending on timescales, perhaps we want a bit of both.

[...]

PR #465, opened 2018-07-30
'Other' Service and Issues

bobwhitelock:
This PR adds an injected 'Other' Service, along with 2 'Other' Issues - one which `requires_service` and one which doesn't. These are the first minimal changes to address https://github.com/alces-software/alces-flight-center/issues/361 - whether and how much these changes improve the situation discussed there remains to be seen.

The core of these changes is the 'Other' Service injection from the experimental branch linked in that issue (thanks @jamesremuscat!); also included is test coverage for this change, fixing of existing tests, the addition of the Issues needed to work with this change, and some other tidying/refactoring to hopefully make this change clearer and more robust.

Trello: https://trello.com/c/EUGNoYqY/410-add-other-service-and-issue-to-facilitate-tier-3-ticket-creation.

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -50,7 +50,7 @@ class CasesController < ApplicationController
     @case = Case.new(my_params.merge(user: current_user))
     authorize @case
 
-    if service_id.present? && service_id.positive?
+    if service_id.present? && service_id.to_i.positive?
       @case.services << Service.find(service_id)
     end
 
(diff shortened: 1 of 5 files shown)


bobwhitelock:
@jamesremuscat I think you're right, if no Components and Services are associated it doesn't mean that the whole Cluster should be associated with the Case, it just means nothing is currently associated. I think it probably shouldn't be possible to create a whole Cluster association from the Case form, since a Case actually effecting a whole Cluster, and so the whole Cluster potentially needing to be brought under maintenance, should be quite a rare event, and admins can later add this association once it is determined it is needed.

I think we may want to remove this line: https://github.com/alces-software/alces-flight-center/blob/3814646fa8159d9c92d1d9a4fb1c413769d113f5/app/models/case.rb#L168 though possibly it's more complex than that, I'm not very familiar with recent changes to this area of the codebase.

PR #468, opened 2018-08-01
change layout of case table

DavidMarchant:
merge 'date' and 'creator' feilds into 'created'
trello: https://trello.com/c/1qEWQGJS/400-collapse-date-and-creator-fields-into-one-field-in-case-table
would you want a hard new line between 'by____' and 'on ____' possibly?

### DIFF
--- a/app/views/partials/_case_table_row.html.erb
+++ b/app/views/partials/_case_table_row.html.erb
@@ -5,10 +5,9 @@
     description:  'Support case created',
     timestamp: kase.created_at
   ) do |content|
-    render 'partials/table_cell_link', url: url, cell_content: content
+    render 'partials/table_cell_link', url: url, cell_content: "By "  + kase.user.name + " on " + content
   end %>
   <td><%= render 'partials/table_cell_link', url: url, cell_content: kase.user_facing_state %></td>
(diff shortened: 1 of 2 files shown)


jamesremuscat:
@DavidMarchant Is this now ready to merge?

DavidMarchant:
Yep, is for me @jamesremuscat

app/views/partials/_case_table_row.html.erb
```diff
@@ -5,10 +5,9 @@
     description:  'Support case created',
     timestamp: kase.created_at
   ) do |content|
-    render 'partials/table_cell_link', url: url, cell_content: content
+    render 'partials/table_cell_link', url: url, cell_content: "By "  + kase.user.name + " on " + content
```
jamesremuscat: Another fun Ruby fact for you: this sort of thing is an ideal use case for Ruby's string interpolation.

Rather than `"foo" + bar + "baz"` you can write `"foo#{bar}baz"` for the same effect, but with better efficiency - the thing inside the hash-bracket is evaluated and inserted into the string.

Note that you need to use double-quotes as strings within single quotes are not interpolated in this way.

DavidMarchant: Ok cheers! I'll edit that and send it up, thanks for the tip

PR #471, opened 2018-08-01
Explicit Case drill down

bobwhitelock:
This PR changes the Case form so that, for each thing which must be selected via a `select` box, the user must explicitly select an option before we show the fields which depend on the value selected.

This is in contrast to the previous behaviour, where the first item in each `select` was initially selected, and forces the user to categorise their problem at each level before proceeding (or at least categorise it as 'uncategorised', via the recently added 'Other' Service and/or Issues).

As can be seen from the size of the diff this ended up being quite involved, and there were also quite a few edge cases to consider. I've also included the removal of the single component/service modes in this PR, as discussed on Slack, which is in the middle of this PR since it both depends on earlier changes and is depended on by later changes.

Have an experiment and let me know what you think about this new UX; hopefully this should provide a little bit more guidance when creating a Case, and make people consider more each decision about how to classify things. It might be nice, if we stick with this Case creation UI, to animate things as different fields appear/disappear (have them slide in or similar), to better indicate what has changed; I've not done this for the moment however since this is beyond the scope of what we want for now.

Trello: https://trello.com/c/7lqtNtRg/413-require-user-to-explicitly-drill-down-and-select-fields-when-creating-case.

This also resolves https://github.com/alces-software/alces-flight-center/issues/361 I believe (we may later want to consider further Case form UI improvements, but I think this should be sufficient for this iteration and we can create a new linked issue with any new information learned once this has been deployed and used for a while).


jamesremuscat:
I'm not going to pretend I understand all of the Elm changes :wink: so the comments I have are from an end-user perspective, are quite minor, and come in no particular order.

- When there is only one option in a dropdown, that option should be preselected. There are now a number of places where your only option is `Other` and it's extra clicks to make that actually selected.
- We're sorting alphabetically at the moment, but the `Other` entries would be better placed at the bottom of their respective dropdowns.
- It feels a bit intimidating to be dumped onto an almost-blank page with a dropdown box demanding you make a decision. I think some friendly help text might make it better. I'll try and come up with some words.
- We should force users who select `Other` to write a more descriptive subject line than the "Other" we generate for them. A case subject of "Other" is never going to be very useful!

If any of these are quick fixes and you've time, please include them in this PR, else we'll add them to the backlog to join those from Mark and Steve once they've had a chance to play.

bobwhitelock:
@jamesremuscat thanks for the review, those are all good points. I've added a new commit to address each of them in turn, as well as some related refactoring in other commits (kept separate so you can see the relevant changes separate from the tidying to make those changes simpler).

Couple of further points related to your comments:

Re:
> - It feels a bit intimidating to be dumped onto an almost-blank page with a dropdown box demanding you make a decision. I think some friendly help text might make it better. I'll try and come up with some words.

I've had a go writing the text for this, but happy for you to adapt if you can think of better wording - if you want to do this it should just require changes here: https://github.com/alces-software/alces-flight-center/blob/777105c22d82de19246f5cb44eac911611e3399d/app/javascript/View/CaseForm.elm#L166-L172

Re:
> - We're sorting alphabetically at the moment, but the `Other` entries would be better placed at the bottom of their respective dropdowns.

and

> - We should force users who select `Other` to write a more descriptive subject line than the "Other" we generate for them. A case subject of "Other" is never going to be very useful!

As discussed IRL yesterday, there's a "flexible to Issues etc. changing names" way to do this and a "hacky but almost certainly fine, and if it stopped working would also be fine" way, and as also discussed I've done the latter. For future reference the logic for this is: if a Service/Category/Issue etc. has a name which starts with 'Other', then shove it down the end of the list, apart from that everything is alphabetically sorted.

Also prompted by you comments, I thought it might be useful if we provided a bit more guidance to users as they gradually select things in the Case form to create a Case. To try and do this I've also added a commit with more descriptive field names for some fields, as well as help text to help let people know what information we need at each step. Let me know what you think to this, and if you want to tweak/add/remove any of these it should just be relatively simple changes either here: https://github.com/alces-software/alces-flight-center/blob/3a5d7bdefd5c50b56e1f3777b2e3ceab5d4b60aa/app/javascript/Field.elm#L67-L80 (for field names), or here: https://github.com/alces-software/alces-flight-center/blob/3a5d7bdefd5c50b56e1f3777b2e3ceab5d4b60aa/app/javascript/Field.elm#L92-L120 (for help text).

app/javascript/View/CaseForm.elm
```diff
@@ -103,38 +148,56 @@ dynamicFieldsSection state =
...
+            """
+            Please attempt to classify your issue as much as possible by
+            selecting from each box in turn, and then provide any further
+            details needed. This helps Alces engineers better determine what
+            problem you are facing and helps us resolve this faster.
+            """
     in
-    if singleCluster then
+    Alert.simpleInfo
+        [ class "d-flex justify-content-between align-items-center" ]
```
jamesremuscat: I don't think these classes are making it through into the markup:

```
<div role="alert" class="alert show alert-info">
  <i class="fa fa-info fa-2x mr-2"></i>
  <span style="flex-grow: 1;">
...
   </span>
</div>
```

bobwhitelock: Hmm, let me check

jamesremuscat: FTR this was a dependency version mismatch - I had a version of `elm-bootstrap` that was too old.

jamesremuscat:
Looks absolutely brilliant now - thanks :smiley_cat:

PR #483, opened 2018-08-07
Consistent date entry fields

atoghill:
Date selection is now consistent for all three areas that require it; viewing check results, depositing credits and making a maintenance request. 

There are three separate definitions for the date picker, one for each use case. While these are not wholly necessary it might be useful for easier individual customisation. However it is also possible to alter many of the date picker options via helper methods.

Another thing to note is that the boxes for selecting the time are unchanged within the maintenance request form as the date picker only supports choosing the date. There is a Trello [card](https://trello.com/c/1yyrqJ63/233-consider-upgrading-date-picker-to-a-date-and-time-picker) for potentially upgrading the date picker to one that can handle both the date and the time.

[Trello](https://trello.com/c/F864pLRC/396-make-date-entry-fields-consistent)

### DIFF
--- a/app/assets/javascripts/datepicker.js
+++ b/app/assets/javascripts/datepicker.js
@@ -1,10 +1,31 @@
 function initialiseDatepicker() {
-  $("#datepicker").datepicker({
+  // Having a datepicker defined for each use case separately isn't necessary
+  // as there are methods you can call to handle changing the disabled days for
+  // example. However for now this offers the ability for easier individual
+  // customisation if we need it.
+
+  $("#restricted-datepicker").datepicker({
     format: "yyyy-mm-dd",
     endDate: '+0d',
     daysOfWeekDisabled: [0,6],
     todayHighlight: true,
     todayBtn: "linked",
+    orientation: "bottom"
+  });
+
+  $("#credit-datepicker").datepicker({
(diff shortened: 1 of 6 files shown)


jamesremuscat:
There's some odd behaviour introduced here when attempting to request a maintenance window that starts in the past. Once you've submitted the page:

 - The datepicker initially has a value of `/` rather than the date that was selected
 - The datepicker is missing the red-bordered validation feedback of the other fields

It's also missing the green-bordered validation feedback that the other fields enjoy.

Please could you investigate?

atoghill:
I'm a bit confused about the green-bordered validation. The current version on `develop` assigns this border to every field regardless of if the date is actually valid or not. But even after submission if the date is valid then the user would never see that page again anyway so assigning the green border after the fact also seems futile. Is there something I am missing here?

jamesremuscat:
Rails magic :wink: 

I care more about the red border than the green; as you say, there's not much value in it since it only gets updated synchronously.

jamesremuscat:
Just had a peek at the docs for `bootstrap-datepicker`. Could we also add a restriction on the maintenance window picker to ensure that the date is no earlier than the current date? It's not possible to request a maintenance window that starts in the past.

PR #489, opened 2018-08-08
Feature/add time to response

### LINKED ISSUE #392: Add time-to-response to credit usage reports
Our current reporting system includes time-to-first-response as a metric against each ticket to demonstrate our operation is within the agreed response times specified in our SLA.

In order to migrate away from the existing manual spreadsheet processes for Flight customers, we need to include this metric in the credit usage reports.

Aside: time-to-resolution might also be nice to have, but it might make the display of the report too congested (however, definitely worth recording from now, or at least having an idea of how we would generate that data retroactively as I think it's something we'll want to display somewhere in the future).  Open to ideas on this.

DavidMarchant:
Trello: https://trello.com/c/99rH87FI/407-add-time-to-response-to-credit-usage-reports
Fixes #392 
Time added with icon, hope the formatting is as desired but can alter as desired
Holiday functionality is yet untested as I'm not inventive enough to think up how that would be done

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,6 +41,7 @@ gem 'state_machines-activerecord',
   ref: 'fed06d9fa64af1cba49d241f4a3ae79626946fe3'
 gem 'state_machines-audit_trail'
 gem 'business_time'
+gem 'holidays'
 gem "audited", "~> 4.7"
 gem 'jwt'
 gem 'pundit'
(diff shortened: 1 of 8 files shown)


DavidMarchant:
Ok, yep, I understand on all fronts, I'll get on that tomorrow

PR #490, opened 2018-08-08
Cancel change request

atoghill:
This PR implements the ability for an admin to cancel a change request that is in the `:draft` state. When a CR is in this draft state it can be cancelled using the 'Cancel' button found on the right-hand panel of the CR page. Once cancelled it should be possible to resolve the case as long as there are no outstanding maintenance requests.

Currently this is the only way to cancel a CR but if necessary it would be relatively simple to add a similar button to the alert displayed above a case.

[Trello](https://trello.com/c/G184hUhN/405-allow-admins-to-cancel-change-requests-from-draft-state)

### DIFF
--- a/app/controllers/change_requests_controller.rb
+++ b/app/controllers/change_requests_controller.rb
@@ -73,6 +73,12 @@ class ChangeRequestsController < ApplicationController
     end
   end
 
+  def cancel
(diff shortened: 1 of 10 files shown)


atoghill:
@jamesremuscat Cheers James, I have just pushed changes to address your comments

Issue #501, opened 2018-08-10
Hide checks tab when no checks are defined


mjtko:
While we're hiding the checks section on the cluster dashboard page when no checks are defined for the cluster, the checks tab is still visible.  Can we hide this too please?

atoghill:
Woops I should have thought about this when I hid the widget in #498. This should be a simple change though 👍

jamesremuscat:
Fixed by #510 (release branch) / #502 (develop).

PR #502, opened 2018-08-10
Hide checks tab when a cluster has no associated checks

atoghill:
When hiding the widget under the same circumstances in #498 I hadn't done the same for the `Checks` tab; see #501. This PR resolves this so that now if a cluster has no cluster checks **both** the widget and the tab are hidden.

[Trello](https://trello.com/c/GRU0oFKz/432-hide-cluster-checks-tab-when-a-cluster-has-no-associated-checks)


jamesremuscat:
Needs merging to both `develop` and `release/2018.6`.

PR #528, opened 2018-08-15
Update text in CR proposed email

### LINKED ISSUE #487: Change to text in "CR proposed" email
Currently it says:

"The change request for this case has been proposed and is awaiting customer authorisation."

As the email is primarily targeted at the customer, I think it would be friendlier to say:

"A change request for this case has been proposed and is awaiting your authorisation."

Atul9:
closes #487

### DIFF
--- a/app/decorators/change_request_state_transition_decorator.rb
+++ b/app/decorators/change_request_state_transition_decorator.rb
@@ -12,7 +12,7 @@ class ChangeRequestStateTransitionDecorator < ApplicationDecorator
   def text_for_event
     case object.event
     when 'propose'
-      'has been proposed and is awaiting customer authorisation.'
+      'has been proposed and is awaiting your authorisation.'
     when 'decline'
       'has been declined.'
     when 'authorise'
--- a/app/models/slack_notifier.rb
+++ b/app/models/slack_notifier.rb
@@ -104,7 +104,7 @@ class SlackNotifier
     end
 
     def change_request_notification(kase, text, user)
-      text = "The change request for this case #{text}"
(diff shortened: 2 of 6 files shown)


mjtko:
Hi @Atul9,

Thanks for your contribution. Could you sign our [Contributor License Agreement](https://www.clahub.com/agreements/alces-software/alces-flight-center) please? Once that's out of the way we can take a look at merging your feature into our development branch for inclusion in the next release!

Atul9:
@mjtko done, I have signed the CLA.

jamesremuscat:
@mjtko Are you happy that this will use "your" in all scenarios, including the case history feed and our internal Slack notifications?

mjtko:
Ah, good catch.

The "The" to "A" change should stand in all places where it's required -- but only for the "propose" action.

It should say "your" in the email but "customer" in the card on the feed (and, less importanly Slack).

Atul9:
@mjtko I checked the flow the text is used in mailer, card and slack notifier, the change in decorator text gets reflected everywhere. Can you please suggest where to update/change for -- `It should say "your" in the email but "customer" in the card on the feed (and, less importanly Slack)`

mjtko:
@Atul9: can't suggest a simple place for an update, cos it doesn't exist yet! :smile:

Unfortunately, updating the `text_for_update` method alone is insufficient to fix the issue, cos the system needs to use a different value depending on the context it's called from.

mjtko:
@Atul9: thanks again for your contribution.  I'm closing this now as the resolution of #487 will require a few more changes than included in this PR.

PR #548, opened 2018-08-23
Import component data from Benchware

### LINKED ISSUE #481: Asset record / component setup and import refactoring
Placeholder for a 2018.7 work, I need @ColonelPanicks to add more detail ;)

1) Import component lists from a cluster script output (benchware?) 
2) Switch the asset record fields to a single markdown page, holding the benchware output at first

jamesremuscat:
This PR fixes #481 by allowing admins to import a cluster's component data from a Benchware YAML export file. We also completely remove the concept of component types, since we have no reliable automatable way of sourcing this information.

A new entry will appear for admins in a cluster's composition dropdown menu (`Cluster` on the cluster dashboard - indeed, now, on any component or service dashboard too) taking them to a simple form where they can upload the YAML file.

The exact specifics of the import process are best described by the included tests; at a high level it is:

```
for each component in file:
    if component already exists in cluster
        update component info field only
    else
        determine or create correct component group
        create new component in group
```

Also included in the tests is a sample file showing the expected format.

Component groups are identified by their new `unix_name`. Some `unix_name`s may automatically be mapped to a friendly name used for display, else the Unix name is used for new groups - though this can be later overridden by an admin.

Trello: https://trello.com/c/inOzPFzQ/447-import-benchware-yaml-component-definitions


jamesremuscat:
Rebased on latest `develop`.

Issue #525, opened 2018-08-15
Split assignee field into two fields


mjtko:
Replace/split the current assignee field into two fields: assigned engineer and assigned site contact.

Only the assigned engineer and assigned site contact can comment on the case.  Assigned site contact is the creator of the case (if a site contact created the case) or the primary site contact (if an admin created the case).  The assigned engineer is the default for the site, as per #413.

Admins can modify engineer assignment and site contact assignment.

If a site contact with open tickets is changed to be a viewer, primary site contact is assigned.

To discuss:

Can site contacts can modify site contact assignment? (To themselves?)
Should an engineer be able to open/withdraw case comments for all site contacts? (Cope with the case assigned site contact is OOO).

ste78:
Can site contacts can modify site contact assignment? (To themselves?)
- I'd say yes, when they are about to go on holiday and want someone else to takeover etc

Should an engineer be able to open/withdraw case comments for all site contacts? (Cope with the case assigned site contact is OOO).
- nice-to-have

jamesremuscat:
@mjtko With reference to #507 - when the assigned contact of a case is changed, who (if anybody) should get an email notification?

mjtko:
Let's send that one out to everybody.

PR #555, opened 2018-08-30
Add assigned contacts to cases

atoghill:
This PR addresses parts of #525. Case assignment has been split into two, with the current field becoming the `Assigned engineer` and the new field the `Assigned contact`. The assigned contact for a case defaults to the creator of the case, or the primary contact for the site if the case is opened by an admin. These values can both be changed by an admin but contacts may only change the assigned contact for a case.

With these changes only those assigned to a case may comment on it. Meaning that at any given time only the assigned engineer and contact will have permission to comment on the case.

Currently notifications follow the same pattern as the current assignee notifications for both engineer and contact assignee changes. @mjtko @jamesremuscat if you feel they should be handled differently then let me know.

screenshot_2018-08-30 alces flight center

[Trello](https://trello.com/c/TUNb8M8V/445-add-site-assignee-to-cases)


app/views/cases/show.html.erb
```diff
@@ -71,9 +71,24 @@
...
+        <td id="case-engineer-assignment">
+          <%= render 'cases/case_assignment_controls',
+            column: :assignee_id,
+            users: @case.potential_assignees,
+            assignee: @case.assignee
+          %>
+        </td>
+      </tr>
+      <tr style="width: 100%;">
+        <td colspan="2"></td>
```
jamesremuscat: It would be a better use of space to instead expand the "Affected components" cell into the whitespace occupied by this empty cell.

app/views/cases/show.html.erb
```diff
@@ -71,9 +71,24 @@
...
+        <th nowrap>Assigned engineer</th>
+        <td id="case-engineer-assignment">
+          <%= render 'cases/case_assignment_controls',
+            column: :assignee_id,
+            users: @case.potential_assignees,
+            assignee: @case.assignee
+          %>
+        </td>
+      </tr>
+      <tr style="width: 100%;">
```
jamesremuscat: I'm not sure this `style` serves any actual purpose.

app/views/cases/_case_assignment_controls.html.erb
```diff
@@ -1,14 +1,14 @@
...
   <%= form_for @case do |f| %>
     <div class="form-group">
       <div class="input-group">
       <%= f.collection_select(
-            :assignee_id,
-            @case.potential_assignees,
+            column,
+            users,
             :id,
             Proc.new { |a| "#{a.admin? ? '* ' : ''}#{a.name}"},
```
jamesremuscat: We don't need to highlight admins with a `*` now that the list for engineer assignees is exclusively admins.

config/routes.rb
```diff
@@ -154,7 +154,7 @@
     end
 
     resources :clusters, only: :show do
-      cases.call(only: [:show, :create, :index, :new]) do
+      cases.call(only: [:show, :create, :index, :new, :update]) do
```
jamesremuscat: With `:update` added here we don't need to define it a second time in the admin routes block (`routes.rb:87`).

spec/controllers/cases_controller_spec.rb
```diff
@@ -290,7 +290,7 @@ def expect_case_created
 
           expect(flash[:error]).to eq "Error updating support case: tier_level cannot be changed when a case is #{state}"
           kase.reload
-          expect(open_case.tier_level).to eq 2
+          expect(kase.tier_level).to eq 2
```
jamesremuscat: Good spot!

spec/features/case/show_spec.rb
```diff
@@ -99,18 +100,15 @@
...
-      expect(event_cards[5].find('.card-body').text).to match(
+      expect(event_cards[6].find('.card-body').text).to eq('First')
+      expect(event_cards[5].find('.card-body').text).to eq('Second')
+      expect(event_cards[4].find('.card-body').text).to match(
         /Maintenance requested for .* from .* until .* by A Scientist; to proceed this maintenance must be confirmed on the cluster dashboard/
       )
-      expect(event_cards[4].find('.card-body').text).to eq 'Changed time worked from 42m to 2h 3m.'
+      expect(event_cards[3].find('.card-body').text).to eq 'Changed time worked from 42m to 2h 3m.'
 
-      expect(event_cards[3].find('.card-body').text).to eq(
```
jamesremuscat: This was a useful test; please reinstate it.

spec/features/case/show_spec.rb
```diff
@@ -360,74 +358,207 @@
...
+        open_case.reload
+
+        expect(open_case.assignee).to eq(user)
+        expect(emails.count).to eq 1
+        expect(emails[0].parts.first.body.raw_source)
+          .to have_text 'Jerry has been set as the assigned engineer for this case'
+
+      end
+
+      it_behaves_like 'contact assignment controls'
```
jamesremuscat: `include_examples` would be a better semantic choice here (and below).

spec/models/case_spec.rb
```diff
@@ -377,22 +377,22 @@
...
     end
 
-    it 'includes any contact for site of Case' do
-      create(:contact, name: 'some_contact', site: kase.site)
+    it 'does not include any user that is not an admin' do
+      create(:viewer, name: 'some_viewer', site: kase.site)
+      create(:contact, name: 'that_contact', site: kase.site)
 
-      expect(subject).to include('some_contact')
+      expect(subject).not_to include('some_viewer' || 'that_contact')
```
jamesremuscat: This doesn't do what you think! This evaluates `'some_viewer' || 'that_contact'`, which of course is `'some_viewer'`, and then performs the `include` test.

A quick look at the source doc for `include` shows that it accepts multiple arguments and will only pass if all arguments are members of the collection e.g.
```ruby
expect([1,2,3]).to include(2,3) # passes
expect([1,2,3]).to include(2,3,4) # fails
```

app/models/case.rb
```diff
@@ -451,6 +465,26 @@ def maybe_set_default_assignee
...
+          la.user = nil
+          la.save!
+        end
+      end
+    end
+  end
+
   def validates_user_assignment
     return if assignee.nil?
     errors.add(:assignee, 'must belong to this site, or be an admin') unless (assignee.site == site) || assignee.admin?
```
jamesremuscat: This validation needs to be updated, and validation for `contact` added.

app/controllers/cases_controller.rb
```diff
@@ -93,36 +93,13 @@ def create
...
-          mailer_method,
-          kase,
-          old_value,
-          kase.send(field)
-        ).deliver_later
-      end
-
-    end
+    fields_changing = params.require(:case)
+      .permit(current_user.admin? ? ADMIN_UPDATABLE_FIELDS : CONTACT_UPDATABLE_FIELDS)
```
jamesremuscat: I'm still wondering if this might be better as two separate methods (and routes) so that Pundit can deal with permissions in the usual way.

app/decorators/audited/audit_decorator.rb
```diff
@@ -43,7 +43,8 @@ def render_card(date, name, type, text, details, admin_only)
     def assignee_id_text(from, to)
       if from
         if to
-          "Changed the assignee of this case from #{User.find(from).name} to #{User.find(to).name}."
+          "Changed the assigned engineer of this case from #{User.find(from).name}"\
+          " to #{User.find(to).name}."
         else
           "Unassigned this case from #{User.find(from).name}."
```
jamesremuscat: These four `else`y cases should also use the "assigned engineer/contact of this case" form text, since otherwise they're indistinguishable across contacts and engineers (without mouseover-ing the icon).

spec/models/case_spec.rb
```diff
@@ -377,22 +377,22 @@
...
+  describe '#potential_contacts' do
+    subject { kase.potential_contacts.map(&:name) }
+    let(:kase) { create(:case) }
 
-    it 'does not include viewer for site of Case' do
-      create(:viewer, name: 'some_viewer', site: kase.site)
+    it 'includes any contact' do
+      create(:contact, site: kase.site, name: 'some_contact')
 
-      expect(subject).not_to include('some_viewer')
```
jamesremuscat: Asserting that this list only includes contacts, not viewers, would be useful. And I hope it doesn't include "any contact" but just the ones for the relevant site...

atoghill:
Rebased on `develop`

db/migrate/20180831145350_initialise_assigned_contacts_for_open_cases.rb
```diff
@@ -0,0 +1,17 @@
...
+    open_cases.each do |kase|
+      user = kase.user
+
+      if user.admin?
+        kase.contact = kase.site.primary_contact
+      else
+        kase.contact = user
+      end
+
+      kase.save
```
jamesremuscat: @mjtko Just checking you're happy that this initial assignment of site contacts will create a card in the case's event history? It won't generate an email or Slack notification as it normally would.

spec/policies/case_policy_spec.rb
```diff
@@ -9,7 +9,7 @@
     it_behaves_like 'it is available only to editors'
   end
 
-  permissions :close?, :assign?, :resolve?, :set_time? do
+  permissions :close?, :assign_engineer?, :resolve?, :set_time? do
```
jamesremuscat: We should have an equivalent check in the test above for the new `:assign_contact?` permission.

db/migrate/20180831145350_initialise_assigned_contacts_for_open_cases.rb
mjtko: Yup, fine for now.  I have a sneaking suspicion we'll need to revisit what cards are created at case creation time (I think one is/my also be created for assigned components), but it's not high up my priority list atm.

atoghill:
Rebased on `develop`

app/views/cases/_case_assignment_controls.html.erb
```diff
@@ -1,14 +1,14 @@
-<% if policy(@case).assign? %>
+<% if policy(@case).assign_engineer? || ( policy(@case).assign_contact? && column == :contact_id ) %>
```
jamesremuscat: Strictly speaking this should be:

```ruby
if (policy(@case).assign_engineer? && column == :assignee_id)
  || (policy(@case).assign_contact? && column == :contact_id)
```
 hence my earlier suggestion to pass the relevant policy result in as a parameter to the template (a pattern we've used elsewhere).

The present code is correct only by virtue of the fact that those with `assign_engineer?` permission will also have `assign_contact?` permission.

spec/models/case/available_support_validator_spec.rb
```diff
@@ -2,6 +2,7 @@
```
jamesremuscat: Think you've accidentally managed to introduce some non-semantic changes to this file.

spec/policies/case_policy_spec.rb
```diff
@@ -5,11 +5,11 @@
 
   let(:record) { create(:case, cluster: create(:cluster, site: site)) }
 
-  permissions :create?, :new?, :escalate? do
+  permissions :create?, :new?,  :escalate?, :assign_contact? do
```
jamesremuscat: You've added some whitespace here!

spec/policies/case_policy_spec.rb
jamesremuscat: This was done :+1:

app/controllers/cases_controller.rb
jamesremuscat: We can consider this in the future, don't worry about it for now.

app/views/cases/_case_assignment_controls.html.erb
atoghill: Yeah that was the reason I hadn't explicitly checked for the column with `assign_engineer?`. If you want it passed in as a parameter I am happy to do that though.

spec/models/case/available_support_validator_spec.rb
atoghill: Yep! Forgot to remove that `,` from when we were trying to fix the issue with this test, woops

PR #560, opened 2018-08-31
Reassign cases when contact is demoted to viewer

atoghill:
When a contact is demoted to viewer any cases assigned to them will be reassigned to the primary contact for the site. For now this will send notifications for each case that has been reassigned but in the future all reassigned cases could be sent in a single list to provide a better user experience.

This is based on #555.

[Trello](https://trello.com/c/XLjGlIRP/446-reassign-cases-when-assigned-contact-becomes-viewer)


atoghill:
Rebased on `develop`

app/models/user.rb
```diff
@@ -90,6 +93,11 @@ def remember_token
     )
   end
 
+  def assigned_cases
```
jamesremuscat: As written, this method:

- Retrieves the association `engineer_cases`
- Retrieves the association `contact_cases`
- Combines them into an array with `|`
- Converts them from an array into an association by querying the database for the records matching the IDs of the records we already just retrieved

I'm assuming the intention here was to provide the set of cases for which either the assigned engineer or the assigned contact is the given user. (If so, `+` might have been clearer than `|`). 

However, since the assigned engineer of a case is necessarily an admin and the assigned contact necessarily _not_ an admin, we already know that (at least) one of those associations will be empty. Therefore, without losing any information, we can simply write

```ruby
engineer_cases || contact_cases
```

which will return the `engineer_cases` association if it's not empty, otherwise the `contact_cases` association, saving the repeated trips to the database of the original method.

app/models/user.rb
```diff
@@ -103,4 +111,20 @@ def site_primary_contact
   def password_optional?
     true  # since we use SSO for passwords
   end
+
+  def validates_case_assignments
```
jamesremuscat: This is a misleading method name, since it does not perform any validation.

Equally, I would expect to see some validation that a viewer has no assigned cases (in which case running the reassignment post-validation is too late).

app/views/case_mailer/reassigned_case.html.erb
```diff
@@ -0,0 +1 @@
+<%= simple_format("This case has been reassigned from #{@old.name} to you as they have been demoted to 'Viewer'.") %>
```
jamesremuscat: I love the use of "demoted" but I think we should use something softer, like `...they have been changed to be a viewer`.

Don't want our customers to start getting competitive :wink:

app/mailers/case_mailer.rb
```diff
@@ -114,6 +114,18 @@ def resolve_case(kase, user)
     SlackNotifier.resolved_case_notification(@case, @user, @text)
   end
 
+  def reassigned_case(kase, old_contact, new_contact)
+    @case = kase
+    @old = old_contact
+    @new = new_contact
+
+    mail(
+      cc: @new&.email,
```
jamesremuscat: In the (valid) case where `@new` is `nil`, it doesn't make any sense to send this email - certainly not as currently worded. It's also worth having a test to document our behaviour in this case.

In the case where `@new` is not `nil`, then we might as well use `to` rather than `cc`, especially as the wording of the email is directed towards an individual.

app/models/slack_notifier.rb
```diff
@@ -171,6 +171,17 @@ def cluster_check_submission_notification(cluster, user, text)
       send_notification(cluster_check_submission_note)
     end
 
+    def reassigned_case_notification(kase, old_contact, new_contact)
+      text = "This case has been reassigned from #{old_contact.name} to #{new_contact.name}"
```
jamesremuscat: This doesn't handle the case where `new_contact` is `nil`.

jamesremuscat:
Rebased on current `develop`.

PR #583, opened 2018-09-05
Feature/order site dashboard

### LINKED ISSUE #435: Ordering of sites on sites dashboard
On the sites dashboard, the ordering of the sites appears arbitrary (presumably whatever order the database/app has decided at some start/restart or other).

Enhancement request is:

1. Introduce a field called something like "site identifier" or "site nick name"
2. Order by this field (alphanumerically descending)
3. Display this field in advance of the official site name on the dashboard (ideally in different style or colour to differentiate)

Reasoning for (1) is so we can both use familiar terms to refer to the site as well as not needing to worry about pronouns or common inversions like "University of X" rather than "X University" etc.

DavidMarchant:
Closes #435 
Can change the table linking to the same type as the case table if that would be desired?
Trello: https://trello.com/c/ABK2xOKx/403-order-sites-on-all-sites-dashboard-by-new-site-identifier-field


app/views/partials/_show_sites.html.erb
```diff
@@ -0,0 +1,21 @@
+<% empty_message = 'No sites found.' %>
```
jamesremuscat: This, and the `_site_table_row` partial, would be better placed in `views/sites` since they're site-specific.

(There's an [outstanding Trello card](https://trello.com/c/MrPpTaLo/234-sort-out-views-partials-directory) to apply the same logic to the existing partials there.)

app/views/partials/_show_sites.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    responsive: false do %>
+    <thead>
+        <tr class="nowrap">
+          <th>Identifier</th>
+          <th>Name</th>
+          <th>Cases</th>
+        </tr>
+    </thead>
+    <tbody>
+      <% sites = sites.sort_by {|site| site.identifier} %>
```
jamesremuscat: Rather than do the sorting here (which is an entirely valid method btw, I'm thinking of future things) I think we should instead set an ordering on the default scope, to make sure that the same ordering is used anywhere else.

That is, add
```ruby
default_scope { order(:identifier) }
```

to `models/site.rb`.

DavidMarchant:
Ok thanks! Will do

jamesremuscat:
Looks good now, thanks!

PR #603, opened 2018-09-06
Set last update for administrative cases to 'N/A'

### LINKED ISSUE #599: Last update for administrative cases should be "N/A"
As we don't send updated to site contacts for administrative cases, when they're in this mode they should display their "Last update" times as "N/A" on the case list rather than "None".

atoghill:
Fixes #599. The `Last update` field for administrative cases now displays as 'N/A'. This uses the same styling as before with 'None' so is in red text, is this something you'd want displayed differently @mjtko?

[Trello](https://trello.com/c/nmZ9ei0u/463-display-n-a-in-the-last-update-field-for-administrative-cases)

### DIFF
--- a/app/views/partials/_case_table_row.html.erb
+++ b/app/views/partials/_case_table_row.html.erb
@@ -21,7 +21,7 @@
   <td><%= render 'partials/table_cell_link', url: url, cell_content: kase.credit_charge&.amount %></td>
   <% if current_user.admin? %>
     <% if kase.last_update.nil?  %>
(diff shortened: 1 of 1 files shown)


jamesremuscat:
Red implies bad, but "N/A" is what we'd expect of an administrative case, so I think we shouldn't be using `text-danger` in that scenario.

mjtko:
I've covered this as part of the case list page design work I've been looking into on and off, so this can be closed as it'll be superseded by that in due course.

PR #614, opened 2018-09-07
Feature/markdown information

DavidMarchant:
Looks like so, any changes welcome: 
image
Couldn't see a view folder that would logically take the new partial so put it in the general one, can move it if desired.
Trello:https://trello.com/c/jbdFP53M/398-add-link-to-markdown-message-that-displays-modal-dialog-describing-basic-markdown-syntax

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -187,23 +187,30 @@ a.filter {
 
 .markdown-modal {
   text-align: center;
-  
-  table {
-    align: center;
 
-    tr {
-      border-top: 1px solid;
-      border-bottom: 1px solid;
-      border-collapse: collapse;
+  .modal-dialog {
+    position: relative;
+    display: table;
+    overflow-x: auto;
+    width: auto;
 
-      ul {
-        float: left;
-        width: 85%;
-      }
+    table {
+      text-align:left;
+      width: 100%;
 
-      ol {
-        float: left;
-        width: 85%;
+      tr {
+        border-top: 1px solid;
+        border-bottom: 1px solid;
+        border-collapse: collapse;
+
+        td {
+          padding: 3px;
+          padding-left: 10px;
+
+          ul {
+            padding-left: 10px;
+          }
+        }
       }
     }
   }
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -189,10 +189,8 @@ a.filter {
   text-align: center;
 
   .modal-dialog {
-    position: relative;
     display: table;
     overflow-x: auto;
-    width: auto;
 
     table {
       text-align:left;
@@ -201,7 +199,6 @@ a.filter {
       tr {
         border-top: 1px solid;
         border-bottom: 1px solid;
-        border-collapse: collapse;
 
         td {
           padding: 3px;
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -184,3 +184,18 @@ a.filter {
     margin-right: -1rem;
   }
 }
+
+.markdown-modal {
+  text-align: center;
+
+  tr {
+    border-top: 1px solid;
+    border-bottom: 1px solid;
+    border-collapse: collapse;
+
+    ul {
+      float: left;
+      width: 85%;
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


mjtko:
The headings look a bit bigger than I'd expect, and the "code" and "blocks of code" styles are the wrong colours -- perhaps the dialog needs a `.markdown` class in it somewhere to make the styles render as the Markdown actually renders? :)

I'd also use "example.com" and "example.com/image.png" for the link and image examples respectively.

Other items that would be good to add if we can fit them in (we might not be able to, and they're not critical) are:

 * Numbered lists: `1. 2.` etc.
 * Horizontal rules: `---`

DavidMarchant:
Ok, yep, will do!

DavidMarchant:
Now looks like 
image
Again any improvements welcome
Decided against including ordered lists in the popup as they didn't really make visual sense - they looked the same in both columns and I felt that those who wanted them would use them regardless of markdown.

mjtko:
LGTM!

app/assets/stylesheets/application.scss
```diff
@@ -184,3 +184,34 @@ a.filter {
     margin-right: -1rem;
   }
 }
+
+.markdown-modal {
+  text-align: center;
+
+  .modal-dialog {
+    position: relative;
+    display: table;
```
jamesremuscat: What is the purpose of `display: table;` here?

As far as I can tell, all it does is break Bootstrap modal positioning.

app/views/partials/_markdown_information.html.erb
```diff
@@ -0,0 +1,89 @@
...
+        This input has Markdown enabled styling<br/>
+        <table class="markdown">
+          <tr>
+            <td><i>Italics</i></td>
+            <td>*text* or _text_</td>
+          </tr>
+          <tr>
+            <td><b>Bold</b></td>
+            <td>**text** or __text__</td>
+          <tr>
```
jamesremuscat: I think you mean `</tr>`.

app/assets/stylesheets/application.scss
```diff
@@ -184,3 +184,34 @@ a.filter {
...
+    width: auto;
+
+    table {
+      text-align:left;
+      width: 100%;
+
+      tr {
+        border-top: 1px solid;
+        border-bottom: 1px solid;
+        border-collapse: collapse;
```
jamesremuscat: This has no effect (`border-collapse: collapse;` is already set for tables).

app/assets/stylesheets/application.scss
```diff
@@ -184,3 +184,34 @@ a.filter {
     margin-right: -1rem;
   }
 }
+
+.markdown-modal {
+  text-align: center;
+
+  .modal-dialog {
+    position: relative;
```
jamesremuscat: `position: relative` and `width: auto;` are already set for Bootstrap's `.modal-dialog` class.

app/views/partials/_markdown_information.html.erb
DavidMarchant: Ah thanks, changed

app/assets/stylesheets/application.scss
DavidMarchant: Ah thanks, removed

app/assets/stylesheets/application.scss
DavidMarchant: Ah thanks, removed

app/assets/stylesheets/application.scss
jamesremuscat: Think you might have missed this comment (Github helpfully decided to hide it).

DavidMarchant: Oh no I didn't, I wrote a response actually but Github helpfully decided to not send it, i'll rewrite it.

The reason it's there is because setting the display property to `table` stops the internal (actual) table from padding unnecessarily to the right to fill the modal. Without it it looks like so: 
image
(compare with the above screenshot)
I'm sure that it isn't the optimum way to do things (it doesn't really make sense) but I couldn't  find another, do you know of one? Apologies for my trial-and-error css

jamesremuscat: If you don't want the table to be full-width, don't tell the table to be full width by giving it `width: 100%`! :wink: 

I don't see anything wrong with that screenshot, to be honest. `.model-dialog` shouldn't have `display: table` because it breaks the modal positioning, so please can we remove it?

DavidMarchant: Yep, absolutely, on it

PR #616, opened 2018-09-10
Automatically expand text areas

### LINKED ISSUE #459: UX: automatically expanding textareas
While the textareas are manually resizable (thanks modern browsers! throwsers.) it would be even better if they automatically increased in height (after a threshold) as they are typed into.

Something like the way the GitHub textareas do it would be ideal.

atoghill:
This PR utilises an `autoresize` script from https://github.com/AndrewDryga/jQuery.Textarea.Autoresize that is available for use under the MIT license. With this script we can automatically expand a `textarea` to match the height of its content.

Fixes #459 when merged.

[Trello](https://trello.com/c/tYdWJhnC/472-automatically-expand-textarea-fields)


atoghill:
@jamesremuscat I couldn't reproduce the text jumping around but I have opted for using the Github variant of autosizing instead now. However switching between `Write` and `Preview` should work fine now so let me know if any of your problems still persist.

jamesremuscat:
I like these very much - you almost don't notice it happening, which is how it should be. Very slick!

I expanded on the tests to explicitly cover the case of entering long text then toggling to preview and back; for reasons unknown this fails on Travis but passes locally, so I've included it in the set of such tests excluded from the Travis test run.

jamesremuscat:
One problem I've just discovered - on the "submit cluster checks" page only the first textarea is given the autoresize blessing.

We probably want to be using `document.querySelectorAll` and iterating over the result - `document.querySelector` will only return the first `textarea` on the page.

atoghill:
Well spotted, the old script I was using handled every textarea on a page. But I forgot to go back and test that fully on the checks page.

app/assets/javascripts/markdown/preview.js
```diff
@@ -42,7 +42,9 @@ function configureMarkdownPreview() {
 function configureEditor(editor) {
   editor.find('[data-markdown-preview-button]').click((e) => showPreview(e, editor));
   editor.find('[data-markdown-write-button]').click((e) => showWrite(e, editor));
-  autosize(document.querySelector('textarea'))
+  document.querySelectorAll('textarea').forEach(function(ta){
```
jamesremuscat: I think here we want to use `editor.find` (plus appropriate changes to the consuming function) else we're reinitialising all the textareas on the page unnecessarily.

atoghill: This should hopefully be resolved as of b67fd17.

app/assets/javascripts/markdown/preview.js
```diff
@@ -42,9 +42,7 @@ function configureMarkdownPreview() {
 function configureEditor(editor) {
   editor.find('[data-markdown-preview-button]').click((e) => showPreview(e, editor));
   editor.find('[data-markdown-write-button]').click((e) => showWrite(e, editor));
-  document.querySelectorAll('textarea').forEach(function(ta){
-    autosize(ta)
-  });
+  autosize(editor.find('textarea').get())
```
jamesremuscat: Textareas no longer auto-expand. This code is throwing exceptions on page load and when preview is toggled. Perhaps I'm getting different behaviour to you but this is now very broken.

atoghill: This should be working now, for both Firefox **and** Chrome at least. Also ensured that this is only being called when we logically expect to see a textarea.

PR #622, opened 2018-09-10
Upload documents to S3 bucket directly

### LINKED ISSUE #492: Document upload
Opening as a placeholder to discuss what we could do in terms of a document upload feature rather than need to upload the wonderfully capable and unlikely to break s3 interface. Basically what sort of timescales would we be looking at to implement something basic that allows me to upload pdf scribbles i've made onsite directly to FC documents page.

jamesremuscat:
This PR enables engineers to upload files directly to the S3 bucket from the cluster documents page.

selection_060

Files are automatically uploaded to the correct bucket and path based on environment/site/cluster. I've modified the `AlcesFlightCenterS3Access` IAM policy to allow Flight Center to `PutObject` within its bucket. As with current files, world-readable permissions are *not* set, nor are they required - we generate a pre-signed URL when making them available through the Documents page.

As with the existing S3 document functionality, all uploaded documents are made available to engineers and to all site users.

Fixes #492.

Trello: https://trello.com/c/3CmfIjpV/464-upload-cluster-documents-to-s3

### DIFF
--- a/app/controllers/clusters_controller.rb
+++ b/app/controllers/clusters_controller.rb
@@ -114,6 +114,9 @@ class ClustersController < ApplicationController
       flash[:error] = 'Could not upload document.'
     end
 
+  rescue Aws::S3::Errors::ServiceError => e
+    flash[:error] = "Could not upload document: #{e.message}"
+  ensure
     redirect_to cluster_documents_path(@cluster)
   end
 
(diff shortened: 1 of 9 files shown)


jamesremuscat:
Note that we'll need to override some default Dokku configuration when we deploy this; I've given details at https://trello.com/c/TuT0BN9Q/477-increase-nginx-max-upload-size.

Issue #579, opened 2018-09-05
Rework/remove "authorisation" message for tier 3 cases during case creation


mjtko:
Primary: remove the warning message displayed when creating a tier 3 case.

Secondary/related:

1. make display of "charging info" available from credit usage page. Rename to: "View support credit allocation basis"
2. make the display of the message during case creation configurable on a cluster-by-cluster basis. Similar rename of link to display modal.

PR #624, opened 2018-09-10
Showing the charging info is now options

benarmston:
Clusters now have a `show_charging_info` flag which is used to control whether a tier 3 case can be created without showing and requiring confirmation of the charging info.


benarmston:
Refs #579

benarmston:
Do you want this rebasing on develop @jamesremuscat?  If so, let me know when you've done reviewing other things and I'll rebase and force push.

jamesremuscat:
Think it's just the schema version so I should be able to handle it - but I'll let you know if not :wink: 

Will also poke Travis into trying the build again, since it seems to have fallen over at the first hurdle.

PR #625, opened 2018-09-11
Fix/indicate hidden case events

### LINKED ISSUE #518: Differentiate between "admin only" and "customer visible" in case event stream
I had forgotten that the "time worked" card (and value) in the case event stream was only visible to admins.  

I think it'd be useful to somehow mark events in the case event stream that are "admin only" so it's clearer what the user is seeing and what admins see.  Different colour or a "padlock" icon or something?

DavidMarchant:
Closes #518 when merged
Currently looks like so 
image
How is this with people?
The icon wasn't my first choice but it seems like many are missing from rails-font-awesome (I thought [user-lock](https://fontawesome.com/icons/user-lock?style=solid) or [user-shield](https://fontawesome.com/icons/user-shield?style=solid) would probably be better)
Also had some issues (again) with the css - the hourglass appears much larger than the `user-secret` to me but both the icons are supposedly 32px tall. This made it fairly difficult to try and dynamically resize them to more similar heights, any ideas?


DavidMarchant:
Ah apologies,  I've created this branch from one of my working branches - not develop. I'll close this and fix it

PR #626, opened 2018-09-11
Add icon to hidden case events

### LINKED ISSUE #518: Differentiate between "admin only" and "customer visible" in case event stream

DavidMarchant:
Closes #518 when merged
Currently looks like so 
image
How is this with people?
The icon wasn't my first choice but it seems like many are missing from rails-font-awesome (I thought [user-lock](https://fontawesome.com/icons/user-lock?style=solid) or [user-shield](https://fontawesome.com/icons/user-shield?style=solid) would probably be better)
Also had some issues (again) with the css - the hourglass appears much larger than the user-secret to me but both the icons are supposedly 32px tall. This made it fairly difficult to try and dynamically resize them to more similar heights, any ideas?

### DIFF
--- a/app/views/cases/_event.html.erb
+++ b/app/views/cases/_event.html.erb
@@ -1,12 +1,10 @@
 <% if !local_assigns.fetch(:admin_only, false) || current_user.admin? %>
   <div class="card event-card">
     <div class="card-header d-inline-flex justify-content-between align-items-center">
-      <span><%= date.to_formatted_s(:long) %> - <b><%= name %></b></span>
       <% if local_assigns.fetch(:admin_only, false) %>
-        <div title="Only visible to engineers">
-          <%= icon('user-secret') =%>
-        </div>
+        <%= icon('lock', {:title=>'Only visible to engineers'}) =%>
       <% end %>
+      <span><%= date.to_formatted_s(:long) %> - <b><%= name %></b></span>
       <div title="<%= details %>">
         <%= icon(type) %>
       </div>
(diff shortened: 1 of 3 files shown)


app/views/cases/_event.html.erb
```diff
@@ -2,8 +2,13 @@
   <div class="card event-card">
     <div class="card-header d-inline-flex justify-content-between align-items-center">
       <span><%= date.to_formatted_s(:long) %> - <b><%= name %></b></span>
+      <% if local_assigns.fetch(:admin_only, false) %>
+        <div title="Only visible to engineers">
```
jamesremuscat: This is a bit of a pointless `div`, its only purpose is to add a `title` to its contents - and I suspect it's the reason you can't get the icons to look the same size.

Add the `title` directly to the FontAwesome icon (something I believe is possible with the `icon()` helper).

DavidMarchant: Ok cheers! I'll look into that

DavidMarchant:
image
How's this for you?

jamesremuscat:
@DavidMarchant Nice!

DavidMarchant:
Great :smiley: can't I just ask before pushing anything is this the best practice for doing to modular css? 

`<div class="<%=local_assigns.fetch(:admin_only, false) ? 'card event-card admin-card' : 'card event-card'%>">`

jamesremuscat:
That's pretty much the pattern we've used elsewhere.

DavidMarchant:
:ok_hand: :ok_hand:

PR #628, opened 2018-09-12
Allow admins to freely change between tiers 2 and 3

### LINKED ISSUE #576: Allow admin to escalate/change tier of case without generating a notification
Additional means to move a case between tiers for admins only.

No emails sent to site contacts, but support team notifications to be generated as usual.

Should generate a card on the case history, tho should state for e.g. "Set this case to tier 3 (General Support)." rather than "Escalated this case to tier 3 (General Support).".

atoghill:
Fixes #576. This PR introduces a new selection box on the 'Tier' row of the case page, only visible to admins. Using this an admin is able to change the tier of a case from within the case page with ease. This generates an event card and sends notifications to the support team but site contacts will not be notified directly.

[Trello](https://trello.com/c/NqJ3xfdA/466-allow-admins-to-freely-set-the-tier-of-a-case)


app/views/cases/show.html.erb
```diff
@@ -135,6 +135,10 @@
           <%= @case.tier_description %>
           <%= render 'cases/escalation_controls' %>
         </td>
+        <td/>
+        <td id="case-tier-assignment">
+          <%= render 'cases/tier_controls' %>
```
jamesremuscat: I know it's a slight deviation from the original spec, but seeing them both together makes me think that we want to be replacing the existing tier description/escalation controls for engineers, rather than adding to them.

It seems odd to offer engineers two methods of achieving the same end goal with slightly different results here - the dropdown, and the yellow "Open for comments" which does exactly the same but also sends a customer-facing email.

@mjtko Would you object if we removed the yellow "Open for comments" button for admins?

jamesremuscat: As with the "escalate" button, we shouldn't be showing this control for administrative cases.

app/mailer_previews/case_mailer_preview.rb
```diff
@@ -46,6 +51,7 @@ def reassign_case
     CaseMailer.reassigned_case(get_case, user, get_case.site.primary_contact)
   end
 
+
```
jamesremuscat: Hello, whitespace...

atoghill: Nothing to see here

app/views/cases/show.html.erb
atoghill: Yeah I wasn't sure on this one either. I left it in as you see it because I knew if you disliked it as well then you would comment on it 👍

app/mailers/case_mailer.rb
```diff
@@ -1,5 +1,6 @@
 require 'slack-notifier'
 class CaseMailer < ApplicationMailer
+  include ApplicationHelper
```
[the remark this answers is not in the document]
atoghill: Aha perfect! Exactly what I was looking for

app/views/cases/show.html.erb
mjtko: Fine by me to drop the "open for comments" button for admins if they have an alternative way of opening for comments (i.e. escalating/setting to tier 3).

jamesremuscat:
Now on tier 4 cases:
- With a non-finalised change request, no tier information is displayed: the 'Tier' cell is empty;
- With a finalised change request, the tier appears to be 2 (because that's the first entry in the select box) rather than 4 (which doesn't appear in the select box).

I would expect:
- With a non-finalised change request, the tier is displayed as "4 (Change request)" with no option to change it
- With a finalised change request, the tier is correctly displayed as "4 (Change request)" in the select box before I change it; once I have changed it and the page reloads, that option should no longer appear

atoghill:
@jamesremuscat I also unintentionally hid the ability for admins to create a CR, so that will be fixed also

atoghill:
@jamesremuscat Back to you!

PR #629, opened 2018-09-13
Fix/hide notes by default

### LINKED ISSUE #601: Default for notes should be engineers only
Just to ensure its an extra click to publish to site contacts.

DavidMarchant:
Changed default to engineers only
Closes #601 when merged
@jamesremuscat when you get a chance to look at this could you confirm if I was correct in changing the controller too please?


app/controllers/notes_controller.rb
```diff
@@ -87,7 +87,7 @@ def cluster_from_params
   def enforce_visibility(params)
     params.tap do |p|
       unless policy(@note).set_visibility?
-        p[:visibility] = 'customer'
+        p[:visibility] = 'engineering'
```
jamesremuscat: The effect of this change would be to force visibility of notes created by users without the `set_visibility?` permission (e.g. site contacts) to `engineering` e.g. customers can create as many notes as they want but they'd instantly vanish as they wouldn't then have permission to see them!

app/views/notes/_form.html.erb
```diff
@@ -33,9 +33,11 @@
           <label for="note_visibility">Visibility</label>
           <%=
             f.select :visibility,
-                     [['Customer and engineer', 'customer'], ['Engineers only', 'engineering']],
-                     {},
-                     class: 'form-control'
+                    options_for_select([['Customer and engineer', 'customer'],
```
jamesremuscat: The effect of this change is that the form will _always_ default to `engineering` visibility _even when editing an existing note with `customer` visibility_.

DavidMarchant:
Ah right, thank you! I was under the impression (for some reason related to internal idiocy) that customers weren't able to add notes, instead only able conditionally view admins' - hence my confusion around `notes_controller.rb:90`!
I'll remove the unnecessary changes shortly - glad I could tangentially be of help in bug finding!

app/views/notes/_form.html.erb
DavidMarchant: Ah yeah, right, I had completely neglected to think about that use case, sorry - changing shortly

jamesremuscat:
Rebased on `develop` to pick up the related tests I've recently added, and added an assertion about default visibility to those.

DavidMarchant:
Ah sorry about the whitespace! :smile: I saw an odd number of spaces and thought something must have gone disastrously wrong

PR #631, opened 2018-09-13
Various (MacOS-centric) development tweaks

mjtko:
Few things that I needed to tweak to smooth over some dev problems on MacOS.  Not that any real developers should ever need to worry about that. Just any pesky fake developers that use MacOs.


spec/models/case_spec.rb
```diff
@@ -392,14 +392,14 @@
...
-      create(:contact, site: kase.site, name: 'another_contact')
+      create(:contact, site: kase.site, name: 'different_contact')
 
       # Viewers and contacts from other sites should not be included
       # in the list of potential contacts
       create(:viewer, site: kase.site, name: 'just_a_viewer')
       create(:contact, site: another_site)
 
-      expect(subject).to eq(['another_contact', 'A Scientist', 'some_contact'])
+      expect(subject).to eq(['A Scientist', 'different_contact', 'some_contact'])
```
jamesremuscat: I'm curious about this one. In what way was this failing for you?

mjtko: On MacOS, the sort order is: `A Scientist`, `another_contact`, `some_contact`

Issue #405, opened 2018-07-05
Consider: UX: highlight case state (with colour?) on case display page


mjtko:
This would give a nice visual indication of the current state of the case.

Issue #400, opened 2018-07-05
Consider: display tier of case on credit usage report


mjtko:
Might be nice to show this as it allows the site contacts to see what tier levels are doing for them.

PR #633, opened 2018-09-13
Redesign of cases table

### LINKED ISSUE #438: Colours for owned tickets are near invisible in certain browsers.
When viewing tickets for a cluster, the tickets owned by oneself are next to invisible. I know on Chrome under MacOS they are relatively visible, however on Chrome 67.0.3396.99 under Linux, the colours are very difficult to tell apart, meaning that the order is somewhat difficult to see! 

See screenshot on Slack: https://alces.slack.com/archives/C72GT476Y/p1532017350000461

### LINKED ISSUE #599: Last update for administrative cases should be "N/A"

mjtko:
This branch includes a number of improvements and changes to the cases table display:

 * Two rows per case:
   * First row contains the case display ID and subject along with its tier (refs #400). Entire row is clickable to view case
   * Second row contains case details and displays a "View" button on the right-hand side on hover. It is no longer all clickable to view case
 * Credit allocation has been dropped from this view as the credit usage page provides that information in a better way
 * Gravatar icons introduced for engineers and contacts
 * Both contact and engineer assignees are displayed (and marked with icons to differentiate)
 * Case state has evolved into a "badge" with an icon and a colour (refs #405)
 * A fuzzy time ago in words badge has been added to the creation date
 * The last update time is now explicitly shown, along with a fuzzy time ago in words badge
 * Creation and update times use "proximity" to determine whether to display simply "Today", "Yesterday", a day of the week (if during the last week) or a full date
 * "My case" highlighting is now performed using a recoloured title row with deeper saturation to make it more visible (fixes #438)
 * Assignees can now be filtered by engineer or contact
 * Assignee filter displays icons to distinguish between engineers and contacts
 * Cases that are assigned to nobody are distinguished with a red exclamation mark "avatar" icon
 * Administrative cases show "N/A" (and a wrench icon) for last update time (fixes #599)
 * Some responsiveness added to improve rendering on narrow displays


jamesremuscat:
A picture is worth 1024 bytes, so:

selection_062

jamesremuscat:
I like where this is going! A handful of comments/questions:

- Rather than the changing `border-top` on each case `tbody`, a `border-bottom` would give better visual separation from the case below in the list - separation from the one above is already provided by the subject/tier bar.
- The almost-black `#111111` isn't a colour we've used elsewhere; I think it'd be nicer to match the Bootstrap `primary` colour we're using throughout the Flight estate (`#2C3E50`, or `#1E2B37` on hover) - or else go the other way, but that would involve a Flight-wide change.
- What's the thinking behind adding Gravatars? Is this a feature you're expecting our customers to be using?
- I'm not a fan of the all-caps `BIZ` without explanation (I think this means '[business time](https://www.youtube.com/watch?v=WGOohBytKTU)'?). Perhaps a tooltip would make this more obvious?
- There's an issue with the display ID box when the background transitions - it briefly flashes white-on-black on mouse out.
- The text and icon for case statuses aren't vertically aligned centrally within the badge.

mjtko:
> Rather than the changing border-top on each case tbody, a border-bottom would give better visual separation from the case below in the list - separation from the one above is already provided by the subject/tier bar.

Not sure I understand what you're getting at -- they look pretty similar to me:

Top:
screen shot 2018-09-13 at 17 50 01

Bottom:
screen shot 2018-09-13 at 17 50 53

Personally I prefer the former (top) one. 😄 

> The almost-black #111111 isn't a colour we've used elsewhere; I think it'd be nicer to match the Bootstrap primary colour we're using throughout the Flight estate (#2C3E50, or #1E2B37 on hover) - or else go the other way, but that would involve a Flight-wide change.

I can only imagine I got it from here https://github.com/alces-software/alces-flight-center/blob/develop/app/assets/stylesheets/cases.scss#L15 but #1E2B37 seems fine!

> What's the thinking behind adding Gravatars? Is this a feature you're expecting our customers to be using?

Main thinking is at-a-glance distinguishing what user the words say.  No expectation -- if they're not/they don't then they'll get a consistently coloured pattern badge thing (very similar to what, for e.g. Slack does).  And if they do want to add their lovely faces, then they're able to. :)

> I'm not a fan of the all-caps BIZ without explanation (I think this means 'business time'?). Perhaps a tooltip would make this more obvious?

Admin-only, wasn't that concerned, just needed a way to denote it wasn't a "standard" duration.  I'm not a big fan of tooltips, let's just add it to our Flight Center Engineer User Guide.

> There's an issue with the display ID box when the background transitions - it briefly flashes white-on-black on mouse out.

I don't see that in my browser, you'll probably have to show me!

> The text and icon for case statuses aren't vertically aligned centrally within the badge.

Good spot!

mjtko:
New commits, addressing review comments:

1. Moved the tbody border (and restyled slightly)
2. Fixed the white-on-black flash
3. Fixed vertical alignment within state badge
4. Use more consistent brand-based colours

app/views/cases/filters/_assigned_to.html.erb
```diff
@@ -1,11 +1,11 @@
-<div class="form-group">
+<div class="form-group assigned-to-filter">
   <label class="form-label" for="filter_assigned_to">Show cases assigned to:</label>
   <select class="form-control" name="assigned_to[]" id="filter_assigned_to" multiple onclick="event.stopPropagation();">
     <% filters[:ranges][:assigned_to].each do |assignee| %>
       <option value="<%= assignee.id %>" <%= filters[:active]["assigned_to"]&.include?(assignee) ? 'selected' : '' %>>
-        <%= assignee.name %>
+        <%= (assignee.admin? ? '&#xf0ad;' : '&#x200a;&#xf007;&#x200a;').html_safe %> <%= assignee.name %>
```
jamesremuscat: What's the intended rendering of these glyphs?

For admins it's a spanner and for Nobody it's a circle-questionmark, but for contacts I get "something that sort of looks like a cropped backtick that disappears when you select that option". (Red circle added for clarity)

selection_063
selection_064

Might be worth pulling these out as constants to make their meaning clear.

jamesremuscat: Possible solution: adding specific classes to the `option`s and then adding style to `::before` with the appropriate Unicode `content` and specifying the FontAwesome font . (I suspect the issue here is that the FA font isn't being applied to the Unicode glyphs, and that the font that _is_ being used 1) differs between Macintosh and Linux, and 2) happens to have glyphs for some of the codepoints being used).

mjtko: Maybe we could try selectively dropping items from/reordering the `font-family` list?

mjtko: Another data point: this renders fine for me on Arch.  Seems like it might be something specific to your distro... Ubuntu?

mjtko: Hmm. Also confirmed this as working on Ubuntu 16.04.  @jamesremuscat: what distro are you using?

jamesremuscat: Ubuntu. Looks like it's whatever font my system is substituting for Helvetica, since when I delete that from the list for `.assigned-to-filter select` it works as intended. Anyone able to try this on Windows, since I suspect that will be more representative of the customer base?

atoghill: Windows test user on the case 👍

atoghill: Looks as intended for me on Windows 10 using Firefox

jamesremuscat: *sigh* I guess I'm outvoted then :wink:

PR #639, opened 2018-09-14
Introduce the concept of engineering levels

### LINKED ISSUE #412: Implement 'Engineering Level' Support
I'd like to be able to add an 'engineer level' for each case logged. 

- levels 1-3
- Only admins can assign a level
- Each admin has a max engineering level stored with their user account
- Tier 1-2 tickets are automatically given level 1 at creation
- Tier 3 tickets are automatically given level 2 at creation
- (UI suggestion) Modify the 'assign to' interface to list names in order of their max level (ideally with a indication of their level)- if name is chosen of a higher level engineer the case level is raised to that level at the same time as it being re-assigned

atoghill:
Both cases and users now have an engineering level that can be assigned to them, between 1 and 3. This value is only applicable to admins/engineers and only they can assign an engineering level to a case. Initially all admins have been given an engineering level of 2 but this can be easily adjusted via the admin interface.

Cases will be assigned an engineering level by default upon creation based on their tier; tiers 1 & 2 = 1, tier 3 = 2. However if an engineer is assigned to a case with a lower engineering level than their own then the case has its engineering level raised to match.

To make this process less surprising the engineering level of an engineer is displayed next to their name on the engineer assignment controls e.g. `(2) Chris Demo`. The names in this list are now also ordered first by their engineering level, in descending order,  and then by their name.

This PR closes #412. 

[Trello](https://trello.com/c/88K1EZ4z/467-add-engineering-level-concept)


jamesremuscat:
Rebased on latest `develop`.

app/views/cases/_engineering_level_controls.html.erb
```diff
@@ -0,0 +1,24 @@
+<% if policy(@case).set_engineering_level? %>
```
jamesremuscat: It's generally better to pass in required parameters to a partial rather than rely on instance variables having been set (e.g. pass `kase: @case` to the `render` call and reference it here as `kase` rather than hoping our caller is in a context with `@case` set).

app/views/cases/show.html.erb
```diff
@@ -139,6 +139,14 @@
           <% end %>
           <%= render 'cases/escalation_controls' %>
         </td>
+        <th>Engineering level</th>
+        <td id="engineering-level-assignment">
+          <% if current_user.admin? %>
```
jamesremuscat: You've got two guards around this data: one testing `current_user.admin?` and one `policy(@case).set_engineering_level?`.

For other case properties it's the partial that decides whether to render the controls or just the value, and it should be using the policy test rather than `admin?`.

app/models/case.rb
```diff
@@ -591,4 +600,15 @@ def handle_tool(type, fields:)
       raise "Unknown type: '#{type}'"
     end
   end
+
+  def set_engineering_level
+    tier_level <= 2 ? self.engineering_level = 1 : self.engineering_level = 2
+  end
+
+  def potentially_raise_engineering_level
```
jamesremuscat: Calling this on a case's `before_validation` works when the case gets updated, but what about when a user's EL gets changed? Cases may be left with incorrect ELs.

db/data/20180913094729_set_engineering_levels_for_cases_and_users.rb
```diff
@@ -0,0 +1,31 @@
...
+      if kase.tier_level <= 2
+        kase.engineering_level = 1
+      else
+        kase.engineering_level = 2
+      end
+
+      kase.save!
+    end
+
+    # Assign EL to admins
```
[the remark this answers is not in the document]
atoghill: Yep apologies I neglected to test the migration again after adding validation in

app/views/cases/_engineering_level_controls.html.erb
atoghill: I'll make sure to change this in any other similar partial on the case page, I know of at least one of the top of my head

atoghill:
@jamesremuscat I have made the relevant changes now

mjtko:
Let's leave that off for now -- open a "Consider:" issue about it tho pls.

app/models/user.rb
```diff
@@ -29,6 +29,15 @@ class User < ApplicationRecord
   }
   validate :validates_primary_contact_assignment
 
+  validates :engineering_level,
+    numericality: {
+      only_integer: true,
+      allow_nil: true,
```
jamesremuscat: We don't want to allow admins to have `nil` as their engineering level.

app/views/case_comments/_form.html.erb
```diff
@@ -1,10 +1,10 @@
-<% if policy(@comment).create? && @case.allowed_to_comment? %>
```
jamesremuscat: Please could these unrelated changes be submitted in a separate PR?

spec/features/case/show_spec.rb
```diff
@@ -977,14 +977,47 @@ def assert_button_successfully_applies(button_text)
               'Set this case to tier 2 (Routine Maintenance).'
             )
           end
+
+          context 'as a contact' do
```
jamesremuscat: This test seems unrelated to this PR?

app/models/case.rb
jamesremuscat: This comment has not been addressed (albeit it's now called `after_update`.

spec/models/case_spec.rb
```diff
@@ -599,4 +606,18 @@
       expect(open_case.administrative?).to eq false
     end
   end
+
+  describe '#potentially_raise_engineering_level' do
```
jamesremuscat: We should also assert that the EL of a case is not decreased when assigned to an engineer with lower EL, and what happens when the case's assignee remains constant but the assignee's EL changes.

spec/features/case/show_spec.rb
atoghill: This test isn't new I just unintentionally moved it to the wrong place. There was some confusing merge conflicts that were split across a few parts of that section of the tests. I'll move it back to the right place.

app/views/case_comments/_form.html.erb
atoghill: The only 'unrelated' changes are in regards to https://github.com/alces-software/alces-flight-center/pull/639#discussion_r217762837. I replied saying I would change any similar cases because it seems illogical to change it in one place and not follow that practice elsewhere. As there were no complaints at the time I wrongly assumed that would be acceptable as it is in relation to the change made for the engineering level controls.

app/models/user.rb
atoghill: Yep well spotted, I neglected to remove that when changing the validation to be conditional

app/models/case.rb
atoghill: Gah I thought I had missed something but when I was scanning through the comments I must have skipped this as it is dealing with the same method as https://github.com/alces-software/alces-flight-center/pull/639#discussion_r217756802.  Apologies

spec/models/case_spec.rb
atoghill: In the specs it states that engineers should have a max engineering level assigned to them so to me that indicates that they shouldn't be assigned to a case with a higher level than their own. However I know that Mark also said not to do any validation between the case and user EL's. Some clarification on this would be wonderful.

PR #640, opened 2018-09-14
Fix/sort component groups

### LINKED ISSUE #588: Order component groups in "All component" list alphanumerically
Currently the ordering looks non-deterministic.  Should be ordered alphanumerically.

This applies both to the groups when displayed on the page as well as in the dropdown menu.

DavidMarchant:
I have, subject to your approval, created a default scope for sorting both the `component groups` and `services` rather than sorting the lists where they are used. I've looked (presumably not exhaustively) for places this otherwise effects the system and the only other place I could see was a cluster's services page but, since now each cluster's component page is similarly sorted, I thought this was probably a benefit. How do you feel about this?
Closes #588 when merged
Trello: https://trello.com/c/aAMsEBsu/490-order-component-groups-in-all-components-list-and-affected-components-picker


jamesremuscat:
Looks good to me, thanks!

PR #642, opened 2018-09-15
Show tiers on credit report (and friends)

### LINKED ISSUE #400: Consider: display tier of case on credit usage report

mjtko:
(Based on #633, cos it reuses some styles.)

Added tier to credit report.  Also, updated "hourglass" button, changed rendering of "time to first response" time to be a bit fuzzier (quarter hours) and added a heading. Various style, wording and iconography tweaks.

Fixes #400


app/views/clusters/_credit_charge_entry.html.erb
```diff
@@ -13,7 +13,14 @@
...
-  <span>
-    <%= pluralize(amount.abs,'credit') %>
+  <% if local_assigns[:tier_level] %>
+     <span class="tier">
+       <span class="case-tier case-tier-<%= tier_level %>">
+         Tier <%= tier_level %>
+       </span>
+     </span>
+  <% end %>
+  <span class="<%= local_assigns[:tier_level] ? "amount-tier-#{tier_level}" : "" %>">
```
jamesremuscat: AFAICT this class doesn't have any effect. Does it have a purpose I've missed?

app/decorators/credit_charge_decorator.rb
```diff
@@ -1,4 +1,11 @@
 class CreditChargeDecorator < ApplicationDecorator
+  TTR_ICON = <<~EOF
```
jamesremuscat: Now that we have column headings, is there much point in keeping these icons?

app/assets/stylesheets/clusters.scss
```diff
@@ -28,23 +28,70 @@ li.credit-charge-entry {
...
+            font-size: 8px;
+            font-weight: bold;
+            top: -8px;
+            font-family: Montserrat;
+        }
+    }
+  }
+
+  .small-text {
+      font-size: 80%;
```
jamesremuscat: I think we can use Bootstrap's `.small` rather than redefine our own here.

app/views/clusters/credit_usage.html.erb
```diff
@@ -85,7 +85,17 @@
       <% if @events.empty? %>
         <p class="text-center no-events-message">No credit usage or accrual in this period.</p>
       <% else %>
-        <ul class="list-group">
+      <ul class="list-group">
```
jamesremuscat: The original intention for the UI here was more "cardy" than "tabley". Adding this header, plus the extra fields we've added since it was conceived, suggests that what we actually want here is a table (and the way you've had to work around this addition in the tests is a further clue that we're fighting the original paradigm).

We should probably bite the bullet and turn this into a boring table rather than an exciting card-driven work of art at some point. :wink:

app/assets/stylesheets/clusters.scss
mjtko: Agreed! TIL!

app/views/clusters/credit_usage.html.erb
mjtko: Yup, let's open a "Consider" issue for it.

app/views/clusters/_credit_charge_entry.html.erb
mjtko: Gone.

app/decorators/credit_charge_decorator.rb
mjtko: Gone.

app/assets/stylesheets/clusters.scss
mjtko: Gone.

app/views/clusters/credit_usage.html.erb
mjtko: Done. (Twice!)

mjtko:
Review comments addressed.

PR #654, opened 2018-09-18
Feature/add charging info to credit usage

### LINKED ISSUE #579: Rework/remove "authorisation" message for tier 3 cases during case creation

DavidMarchant:
Closes #579 when merged (I think? I'm told part 2 has been done)
Instead of attempting to interface with the existing elm modal for the same purpose (which made my head hurt) I recreated it as its own partial - will this be an issue because of repeating code? Tell me if so & I'll persevere with elm.
The two look functionally identical.
Also how's the placement of the button for you? 
image
Can move it if desired ofc.


app/views/clusters/_charging_info_modal.html.erb
```diff
@@ -0,0 +1,26 @@
+<button class="btn btn-primary header-btn" data-toggle="modal" data-target="#charging-modal"><%=label%></button>
+<div class="modal fade charging-modal" id="charging-modal" role="dialog">
+  <div class="modal-dialog modal-dialog-centered" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title"><%= cluster + " charging info" %> </h5>
+        <button class="close">x</button>
+      </div>
+      <div class="modal-body">
+        <p><i>Credit allocation to cases is based on the total time worked by our engineers:</i></p>
```
jamesremuscat: Rather than being hard-coded, this should use each cluster's configured charging information (e.g. `cluster.charging_info`), rendered as Markdown.

You can add
```ruby
include MarkdownColumn(:charging_info)
```
in the `Cluster` model, and then call `cluster.rendered_charging_info` to get the properly-rendered version.

app/views/clusters/credit_usage.html.erb
```diff
@@ -2,7 +2,11 @@
 <%= render 'partials/tabs', activate: :credit_usage do %>
   <div class="card-body row">
     <div class="col-sm-12">
-      <h3><%= cluster.name %> support credit usage</h3>
+      <h3 class="credit-usage-heading"><%= cluster.name %> support credit usage</h3>
+      <%= render "clusters/charging_info_modal",
+          label: "View support credit allocation basis",
```
jamesremuscat: We never give this modal a different value for `label` - is there a particular reason it's a parameter to the partial?

app/views/clusters/_charging_info_modal.html.erb
```diff
@@ -0,0 +1,26 @@
+<button class="btn btn-primary header-btn" data-toggle="modal" data-target="#charging-modal"><%=label%></button>
+<div class="modal fade charging-modal" id="charging-modal" role="dialog">
+  <div class="modal-dialog modal-dialog-centered" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title"><%= cluster + " charging info" %> </h5>
```
jamesremuscat: "support credit allocation basis" rather than "charging info" please.

app/views/clusters/_charging_info_modal.html.erb
```diff
@@ -0,0 +1,26 @@
...
+          <li><p>30 minutes to 3 hours ➝ <b>1</b> credit</p></li>
+          <li><p>3 to 6 hours ➝ <b>2</b> credits</p></li>
+          <li><p>6 to 9 hours ➝ <b>3</b> credits</p></li>
+          <li><p>9 to 12 hours ➝ <b>4</b> credits</p></li>
+          <li><p>Over 12 hours ➝ <b>5</b> credits</p></li>
+        </ul>
+        <p><i>Note that time worked is measured in business (not elapsed) hours.</i></p>
+      </div>
+      <div class="modal-footer">
+        <button class="btn btn-outline-primary">Close</button>
```
jamesremuscat: The `Close` button doesn't.

app/views/clusters/_charging_info_modal.html.erb
```diff
@@ -0,0 +1,26 @@
+<button class="btn btn-primary header-btn" data-toggle="modal" data-target="#charging-modal"><%=label%></button>
+<div class="modal fade charging-modal" id="charging-modal" role="dialog">
+  <div class="modal-dialog modal-dialog-centered" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title"><%= cluster + " charging info" %> </h5>
+        <button class="close">x</button>
```
jamesremuscat: This close button doesn't, either.

app/views/clusters/_charging_info_modal.html.erb
DavidMarchant: Aha, right you are, idiotic of me not to test that, i always just click outside i suppose

app/views/clusters/_charging_info_modal.html.erb
DavidMarchant: As with the other, sorry

app/views/clusters/_charging_info_modal.html.erb
DavidMarchant: sure thing

app/views/clusters/credit_usage.html.erb
DavidMarchant: Umm just possible code re-use elsewhere at somepoint I suppose, should I change it?

mjtko:
In terms of positioning/styling, can we render this as a link instead of a button and place it under the service plan pull please, e.g.:

screen shot 2018-09-18 at 16 28 21

Also happy to drop the "support" bit from the copy, cos I think it's obvious it's referring to support credits in this context.

mjtko:
For reference, this was the markup I used for the mockup:

```html
<span>
        <a href="#" style="
    text-decoration: underline dotted;
    font-size: 85%;
    "><i class="fa fa-eye"></i> View credit allocation basis
</a>
    </span>
```

app/views/clusters/_charging_info_modal.html.erb
DavidMarchant: Ah right ok! didn't realise that the data was variable, sorry, I'll change that & shout if i have any issues

DavidMarchant:
Ok, sure, cheers @mjtko!

DavidMarchant:
Ok James, they've all been made - I've removed the label param in favour of a static label assuming that was better in your eyes :slightly_smiling_face:

DavidMarchant:
Ah right, sorry!

PR #655, opened 2018-09-18
Change request workflow alterations

### LINKED ISSUE #504: Change request workflow enhancements (was: Allow admin to reopen change request after cancellation)
Another workflow enhancement -- if a change request is cancelled, then it turns out a CR is actually needed, we're currently stuck in the terminal "cancelled" state.

Can we add a "reopen" button (for admins only) to put the CR back into draft?

atoghill:
This PR fixes #504. Admins have access to every button during the change request process now as well as some new ones to handle the following transitions:

- Cancelled -> Draft
- Awaiting authorisation -> Draft
- Declined -> Draft
- Handover -> In progress

There are aspects of these changes I am still unsure of but making changes should be simple now that everything is in place.

[Trello](https://trello.com/c/SqYosyqe/468-cr-workflow-alterations)


app/decorators/change_request_state_transition_decorator.rb
```diff
@@ -25,6 +25,8 @@ def text_for_event
       'has been cancelled.'
     when 'request_changes'
       'has had further changes requested.'
+    when 'retract'
+      'has been retracted and changes are being made.'
     end
```
jamesremuscat: We're missing a case for `reopen` here.

atoghill:
Yep you hit on most of the things I was unsure of so I'll wait for a reply before resolving these

mjtko:
> the wording used when moving a CR from proposed back to draft: `Change request XXX99 has been retracted and changes are being made.`

"Retracted" is possible a bit strong.  Maybe `Change request XXX99 has been placed back into draft so changes can be made.`

> that doing so should send an email to the customer

Yup, seems fair enough as they've received an email saying it's been proposed, so an email telling them that it has been unproposed seems fine. :)

> likewise, that reopening a change request (from declined or cancelled back to draft) should send an email to the customer.

I don't think we send an email when we initially open a CR (i.e. escalate the case to tier 4 and start drafting), so I don't think this needs an email either.

atoghill:
Just a note the wording for that was generalised as there are two different transitions that the `retract` event handles; awaiting_authorisation -> draft, in_handover -> in_progress. This doesn't change the fact I'll reword things but it is important to know it wasn't just for that transition.

jamesremuscat:
@atoghill Think you might have missed one of my comments:

> It doesn't make any sense to allow reopening of a CR unless its case is still open, so we should disallow that.

PR #660, opened 2018-09-19
Render state badge on case view page. Fixes #405.

### LINKED ISSUE #405: Consider: UX: highlight case state (with colour?) on case display page

mjtko:
Ronseal. Fixes #405.


mjtko:
:+1: for moving to partial.  Done that and rebased on develop to avoid conflicts with tonkers-code.

PR #664, opened 2018-09-19
Update this test to match changes to SlackNotifier.

jamesremuscat:
As title.

(Note for interested observers: this is one of the tests that consistently passes locally but consistently fails on Travis, hence is disabled for CI.)


mjtko:
Ah, my bad, I didn't run the `js` tests cos they don't work for me out of the box and I didn't dig into any further, so I didn't spot this one needed changing.  Thanks for fixing!

PR #671, opened 2018-09-20
Count case creation as an update

### LINKED ISSUE #597: If admin opens a case, time since last update should consider that an update
https://github.com/alces-software/alces-flight-center/issues/479#issuecomment-412938079

Our current implementation assumes "no", "no" answers to the questions in the above comment.  We should update this to "yes", "no", as in:

If an admin opens a case, the time since last update should use the time the case was opened, as it was the last time a message was sent to the customer regarding the case.

atoghill:
Fixes #597. Upon case creation `last_update` is set to the current time.

[Trello](https://trello.com/c/0uoIUume/503-a-cases-lastupdated-should-be-the-time-it-was-opened-when-raised-by-an-admin)


atoghill:
@jamesremuscat Apologies! I did not test this thoroughly enough. This should now only set `last_update` upon case creation when the user is an admin. I have included some better tests that check the interaction from both sides now.

app/mailers/case_mailer.rb
```diff
@@ -10,7 +10,7 @@ def new_case(my_case)
     mail(
       cc: @case.email_recipients,
       subject: @case.email_subject,
-      bypass_timestamp_update: true
+      bypass_timestamp_update: @case.user.admin? ? false : true
```
jamesremuscat: Or even, `!@case.user.admin?` :slightly_smiling_face:

PR #678, opened 2018-09-21
Bulleted list of components in maintenance emails

### LINKED ISSUE #529: Bullet list of components in maintenance request email
Readability of the maintenance request email when a request is made for many components could do with a little improvement.

i.e. go from something like this:

> Maintenance requested for Hardware Management (service), node802 (component), node803
> (component), node919 (component), node920 (component), node921 (component), node922 
> (component), node923 (component), node924 (component), node925 (component), node926 (component)

To something like this:

> Maintenance requested for:
>  * Hardware Management (service)
>  * node802 (component)
>  * node803 (component)
>  * node919 (component)
>  * node920 (component)
>  * node921 (component)
>  * node922 (component)
>  * node923 (component)
>  * node924 (component)
>  * node925 (component)
>  * node926 (component)

Not sure if it needs to say "component" and "service" -- those seem pretty self-explanatory from the naming of the item in any case but I'm open to being convinced otherwise of this. 😄

atoghill:
This PR implements the initial pass of adding bullet points to maintenance emails. Whilst this does do what it says I am not completely happy with how the notifications look and I'm sure you'll have thoughts on this.

Once this PR is merged it fixes #529.

[Trello](https://trello.com/c/9h0nwsX6/504-bulleted-list-of-components-in-maintenance-email)


mjtko:
Can't we just use Markdown for this? i.e. render the text with ` * bullet point` stuff, and then pass through the Markdown renderer for HTML emails?

atoghill:
We could use Markdown for this and I'll admit that thought hadn't crossed my mind. But I don't think it would make this process any easier as we'd still need to handle the Slack notifications differently like I have here. In a perfect world everyone would use the same syntax and we wouldn't need to special case Markdown related things.

mjtko:
Don't see why Slack notifications require special handling -- they should show the markdownesque `*` chars fine as bullets.

atoghill:
They would unless there was only one component. As far as I can tell if you have a singular `* word` no bullet point is created.

mjtko:
What I mean is output using `*` in Slack is fine (as it also is in the plain text email):

```
Things under maintenance:

 * a thing
 * another thing

End.
```

mjtko:
I mean do this kind of thing:

https://github.com/alces-software/alces-flight-center/blob/c7974ed43ae6b416dbef5d6cb07bd78d84aef7bb/app/views/case_mailer/comment.html.erb#L6

Here:

https://github.com/alces-software/alces-flight-center/blob/c7974ed43ae6b416dbef5d6cb07bd78d84aef7bb/app/views/case_mailer/maintenance_ending_soon.html.erb#L1

And here:

https://github.com/alces-software/alces-flight-center/blob/c7974ed43ae6b416dbef5d6cb07bd78d84aef7bb/app/views/case_mailer/maintenance_state_transition.html.erb#L1

And change the `@text` passed in here:

https://github.com/alces-software/alces-flight-center/blob/c7974ed43ae6b416dbef5d6cb07bd78d84aef7bb/app/mailers/case_mailer.rb#L72

Probably by modifying how the text is constructed in https://github.com/alces-software/alces-flight-center/blob/develop/app/models/maintenance_notifier.rb

mjtko:
Something like this comes to mind:

```ruby
def model_names_as_bullets
  associated_model_names.map {|n| " * #{n}"}.join("\n")
end
```

And then update the text constructed in each of the methods to use that, for e.g.:

```ruby
  def request_comment
    <<~EOF
      Maintenance requested for:

      #{model_names_as_bullets}

      Maintenance scheduled from #{requested_start} until #{expected_end} by #{window.requested_by.name}.

      To proceed, this maintenance period must be confirmed on the cluster dashboard:

      #{cluster_dashboard_url}
    EOF
  end
```

There are several other messages that will need restructuring/rewording.  Give me a shout if you would like me to look at what you have come up with/come up with it.

I think just drop the call to `squish` entirely: https://github.com/alces-software/alces-flight-center/blob/c7974ed43ae6b416dbef5d6cb07bd78d84aef7bb/app/models/maintenance_notifier.rb#L5

Ideally we'd do something clever with the cluster dashboard URL, but anything that works in Markdown looks unpleasant in plain text, so let's stick with just leaving the URL inline as is for the moment.

jamesremuscat:
Closing in favour of #700.

PR #696, opened 2018-09-24
Add code fence block syntax highlighting

mjtko:
Forgot to open this one!

This adds syntax highlighting and line numbering for code fence blocks when started using the "```\<lang>" style, allowing us to do snazzy things like:

---
\```bash
#!/bin/bash
echo "I am syntax highlighted!"
\```

---

and:

---
\```shell_session
$ ls -al /
things
listed
here
\```

---

Which render something like this:

screen shot 2018-09-24 at 15 20 26


jamesremuscat:
Just an observation that the rendering of this sort of markup in Slack notifications is unchanged but still probably the best we can do with Slack's very limited vocabulary:

selection_001

PR #708, opened 2018-09-26
Fix case filtering on admin landing page

### LINKED ISSUE #607: The filters on the global dashboard view yield unexpected results
Trying to filter by anybody other than yourself on the global dashboard view, or by any state other than "open" yields no results.  Using these filters should either not be possible from the global dashboard view, or take you to the cases view when you've set them.

atoghill:
This PR fixes #607, allowing for successful filtering of cases on the admin landing page (`My Cases` view).

[Trello](https://trello.com/c/ko0TJGg8/520-fix-filtering-from-the-my-cases-view-of-the-global-dashboard)


atoghill:
With #715 I have adjusted the subtitle for case pages to reflect that cases may be split across various pages. Therefore it may be that we do away with the old "...for current_user.name" format entirely.

Yeah due to the nature of the system when it clears the filters it reloads the page, at which point the default setting for the `My Cases` page is to display all active cases for the current user. So it depends how we view that page. If we see that page as a filter on top of all cases then it should definitely not go back to the defaults for that page. However if we see it as a page where filters can be applied on the set of cases for the current user then I think the current behaviour is acceptable as you're simply returning to the `My Cases` view.

I also noticed that last issue when viewing a page with no available cases.

atoghill:
Rebased on `develop`

app/views/cases/filters/_filter_for.html.erb
```diff
@@ -15,7 +15,7 @@
       <input class="btn btn-primary btn-sm" type="submit" value="Apply"/>
       <%=
         link_to 'Clear all filters',
-                request.path,
+                cases_path,
```
jamesremuscat: This changes existing behaviour: the problem with this implementation is that, as well as clearing filters, it also changes the current scope to the most top-level available to the current user (`AllSites` for admins, or a particular `Site` for site users).

I think the behaviour of the "clear all filters" button should be:

- In the `AllSites` scope: go to `/cases` and show all cases
- In a site scope: go to `/sites/<site>/cases` (which ought to be the current URL anyway) and show all cases (retain existing behaviour)
- In a cluster scope or below: go to `/clusters/<cluster>/cases` (which ought to be the current URL anyway) and show all cases (retain existing behaviour)

Justification: the scope of the list isn't a filter, so shouldn't be changed by clearing filters.

atoghill: 2f51e3e hopefully satisfies this

PR #715, opened 2018-09-28
Case pagination

### LINKED ISSUE #446: Paginate cases list
The resolved cases list is starting to take a long time to load.

We should implement pagination to reduce the amount of data needed to be retrieved to render the page (and thus the time it takes to load/render).

atoghill:
The long awaited paginating of cases has arrived and with it comes (somewhat) quicker case page loading times. There are still optimisations that could be done to improve the loading times further but they are unrelated to pagination. Paginating the case list alone has improved loading times from ~12000ms to ~3000ms for myself locally on average when loading all cases.

Closes #446 when merged.

#### Page entry display

screenshot_2018-09-28 cases - global - alces flight center

#### Page Controls
screenshot_2018-09-28 cases - global - alces flight center 1

[Trello](https://trello.com/c/UY6p2laA/524-pagination-of-case-lists)


app/views/partials/_show_cases.html.erb
```diff
@@ -30,4 +30,8 @@
       <% end %>
     </tbody>
   <% end %>
+  <hr>
+  <div style="justify-content: center;">
```
jamesremuscat: Not sure that this `style` is having any effect; it's the `.justify-content-center` on `ul.pagination` that's centring the paginator (for it is that element that has `display: flex`).

app/views/partials/_show_cases.html.erb
```diff
@@ -30,4 +30,8 @@
...
+    <%= paginate cases, window: 2 %>
```
jamesremuscat: I wonder if we should also show a paginator at the top of the list. What do you think?

app/views/partials/_show_cases.html.erb
atoghill: Yep you're right, this is me being a fool and not removing that style like I thought I had.

app/views/partials/_show_cases.html.erb
atoghill: If you think it would be worth having at the top then I could definitely throw one there

atoghill:
@jamesremuscat Would you prefer the paginators to be outside of the form area?

jamesremuscat:
I think they're fine where you've put them - but I would use something like an `mt-2` class on the paginator `nav` (or `ul`) rather than have a `<br/>` before it, since that adds extra whitespace in the case where the paginator is not shown.

atoghill:
The last commit has moved them outside of the main form and I personally think it looks a lot better here, let me know what you think

jamesremuscat:
LGTM, thanks!

Issue #538, opened 2018-08-17
Ability to add multiple components to a log entry.


RuanEllis:
Just like we can with cases now, would it be possible in future to add multiple components/services to a log entry? We can currently only add single components to a log entry, but often the actions are performed on multiple nodes.

jamesremuscat:
Related to #539.

PR #724, opened 2018-10-02
Allow logs to be associated with multiple components

atoghill:
Part 1 of #538. This PR is the backend work for allowing logs to have multiple components associated with them. The process of creating logs is no different from before however future work will add UI for associating multiple components to a log.

[Trello](https://trello.com/c/u7HGmH9y/469-allow-log-entries-to-be-associated-with-multiple-components)


app/controllers/logs_controller.rb
```diff
@@ -2,16 +2,35 @@
 class LogsController < ApplicationController
   def index
     @scope = @component || @cluster
-    @new_log = Log.new(@scope.class.name.downcase => @scope)
+    @new_log = if @scope.is_a? Component
+                 Log.new(components: [@scope])
+               else
+                 Log.new(@scope.class.name.downcase => @scope)
```
jamesremuscat: As discussed, you could eliminate this conditional by using

```ruby
@new_log = Log.new(@scope.class.name.downcase.pluralize => [@scope])
```

app/controllers/logs_controller.rb
```diff
@@ -2,16 +2,35 @@
...
     @components = @scope.components if @scope.is_a? Cluster
   end
 
   def create
     @scope = @component || @cluster # TODO: Generalise this
-    new_log = @scope.logs.build(log_params)
+
+    my_params = log_params
+
+    component_id = my_params.delete(:components)
```
jamesremuscat: Bit confused by this. Is `component_id` a component ID? Is `params[:components]` a list of components? Something might need renaming to make this more obvious.

app/controllers/logs_controller.rb
```diff
@@ -2,16 +2,35 @@
...
+
+    new_log = @scope.logs.build(my_params)
     authorize new_log
+
+    if component_id.present?
+      new_log.components << Component.find(component_id)
+    end
+
+    if @scope.is_a? Component
+      new_log.components << Component.find(@scope.id)
```
jamesremuscat: But you already _know_ that `@scope` is the `Component` you're looking for... it's right there in the line above!

app/mailers/admin_mailer.rb
```diff
@@ -14,7 +14,7 @@ def daily_open_cases_list(admin, case_ids)
   def new_log(log)
     @log = log
     @text = "New log created on #{@log.cluster.name}" \
-      " #{@log&.component ? 'for ' + @log.component.name : nil }"
+      " #{@log&.components ? 'for ' + @log.decorate.associated_components : nil }"
```
jamesremuscat: We should not repeat history unnecessarily, and so this should be a bulleted list of associated things.

app/views/logs/index.html.erb
```diff
@@ -10,16 +10,16 @@
...
     </thead>
     <% @logs.each do |log| %>
       <tr>
         <%= timestamp_td(description:  'Logged', timestamp: log.created_at) %>
         <td class="nowrap"><%= log.engineer.name %></td>
         <td><%= log.rendered_details %></td>
-        <td>
-          <%= log.component&.decorate&.link || raw("<em>None</em>") %>
+        <td class="d-none d-lg-table-cell">
+          <%= render 'partials/association_summary', associations: log.components %>
```
jamesremuscat: Why restrict this to `log.components` rather than `log.associations`?

app/models/log.rb
```diff
@@ -6,11 +6,23 @@ class Log < ApplicationRecord
...
   has_and_belongs_to_many :cases
 
+  has_many :log_associations, dependent: :destroy
+  has_many :components,
+           dependent: :destroy,
+           through: :log_associations,
+           source: :associated_element,
+           source_type: 'Component'
+
+  has_many :component_groups,
```
jamesremuscat: The original request in #538 did also include `Service`s in the list of "things that could be associated with a log".

app/mailers/admin_mailer.rb
jamesremuscat: Why limit this to just components rather than all associated things? (nb it's fine to use the terminology "Associated components" in the UI as 1) we do elsewhere 2) there's not really a better word)

app/models/slack_notifier.rb
```diff
@@ -87,9 +87,9 @@ def log_notification(log, text)
 
         a << {
           title: 'Component',
-          value: log.component.name,
+          value: log.decorate.associated_components,
```
jamesremuscat: Again - why limit this to just components?

app/models/log.rb
```diff
@@ -37,14 +49,17 @@ def cases_belong_to_cluster
...
+    components.each do |c|
+      correct_cluster = (c.cluster == cluster)
+      errors.add :component, COMPONENT_CLUSTER_ERROR unless correct_cluster
+    end
   end
 
   def assign_cluster_if_necessary
     return if cluster
-    self.cluster = component.cluster if component
+    component = components.first
```
jamesremuscat: Why not `associations` etc?

app/models/log.rb
```diff
@@ -37,14 +49,17 @@ def cases_belong_to_cluster
 
   COMPONENT_CLUSTER_ERROR = 'must be in the same cluster as the log'
   def component_belongs_to_cluster
```
jamesremuscat: This method should consider not just components, but all associations (and hence get renamed too).

app/controllers/logs_controller.rb
atoghill: My only issue with that was that when `@scope` is a cluster it ceases to function

jamesremuscat: Only if you don't have `clusters` as a possible association (both cases and maintenance windows do). The purpose of that is to associate with an entire cluster.

app/controllers/logs_controller.rb
atoghill: Yeah I wasn't entirely sure when to change this one as currently you can only select one component. However in a future PR this would definitely need to be adjusted, I can make that change preemptively if you think that would be better.

app/controllers/logs_controller.rb
atoghill: Are you referring to the `if` statement above it? When you're within the context of a `Component` there is no component sent in `params` hence why I added this. Alternatively I could try and ensure the component is sent but this seemed like a viable solution at the time.

jamesremuscat: The function of this piece of code is not in question; its method of achieving that function is.

This code first checks if `@scope` is a component. If it is, then it looks up the component that has the same ID as `@scope` - which, since IDs are unique, turns out to be `@scope` after all.

You don't need to do that extra database lookup, since if `@scope.is_a? Component` then `@scope` *is* the `Component` with ID `@scope.id`!

app/models/log.rb
atoghill: As of https://github.com/alces-software/alces-flight-center/pull/724/commits/950ce16f27fcf8556df68429bc30eb164799a2a1 services are linked with logs

jamesremuscat:
Rebased on latest `develop`.

app/controllers/logs_controller.rb
```diff
@@ -2,16 +2,31 @@
...
     @components = @scope.components if @scope.is_a? Cluster
   end
 
   def create
     @scope = @component || @cluster # TODO: Generalise this
-    new_log = @scope.logs.build(log_params)
+
+    my_params = log_params
+
+    components = my_params.delete(:components)
```
jamesremuscat: Calling this `components` implies that it's a multitude of `Component`s, when really - for the moment at least - it's an optional single component's ID...

I suspect you'll be immediately changing this in your next PR, though, so on that assumption no point doing so here.

app/controllers/logs_controller.rb
```diff
@@ -2,16 +2,31 @@
...
+    new_log = @scope.logs.build(my_params)
     authorize new_log
+
+    if components.present?
+      new_log.components << Component.find(components)
+    end
+
+    if @scope.is_a? Component
+      new_log.components << @scope
+      new_log.cluster = @scope.cluster
```
jamesremuscat: Isn't this the sort of thing (assigning `cluster` to a `Log`) that ought to be handled by `Log#assign_cluster_if_necessary`? (It only isn't because it's called `after_initialize` rather than at another point in the model lifecycle.)

app/models/slack_notifier.rb
```diff
@@ -87,9 +87,9 @@ def log_notification(log, text)
 
         a << {
           title: 'Component',
```
jamesremuscat: Perhaps `Component(s)`?

app/models/log.rb
```diff
@@ -6,18 +6,51 @@ class Log < ApplicationRecord
   default_scope { order(created_at: :desc) }
 
   belongs_to :cluster
-  belongs_to :component, optional: true
   has_one :site, through: :cluster
   belongs_to :engineer, class_name: 'User', foreign_key: 'user_id'
   has_and_belongs_to_many :cases
 
+  has_many :log_associations, dependent: :destroy
+  has_many :services,
```
jamesremuscat: Note that you've not (yet) included the reverse relationships so there's no way to navigate from a `Service` model to logs related to that model. Likewise for component groups. You might have been planning to do this in the next PR; I'll add a card on Trello for adding the relevant UI pieces to view logs for a service or component group in the same way we can for components.

app/controllers/logs_controller.rb
atoghill: Yeah I made a pre-emptive change because of your previous comment about it being called `component_id` before 😛

app/models/slack_notifier.rb
atoghill: It seems that a fair few of your comments are regarding changes that I intended to do in the next PR when it is actually possible to create a log with multiple components. I am more than happy to do all of these changes in this PR so that the next PR is just the UI.

atoghill:
@jamesremuscat The rest of your comments are to be addressed in another PR soon™

PR #734, opened 2018-10-04
Make Case form usable when no Components available for current Cluster

### LINKED ISSUE #731: Cannot create any cases when a cluster has no components
When a cluster has no components, the case creation form refuses to render.

The set up process is currently working like this:

1. Create the site (if it doesn't yet exist)
2. Create the cluster with basic information
3. Create change request cases

The hardware list may or may not have been imported by the time we need to perform (3), so the cases form needs to operate without components being created.

bobwhitelock:
Fixes #731 (https://trello.com/c/PWoAgtc6/539-cannot-create-any-cases-when-a-cluster-has-no-components).

This is done by:

- allowing the Case form to successfully load when any Cluster has no Components (previously it failed to initialise as the `components` field for the Cluster would be in an unexpected format);

- when a Cluster has no Components, preventing selection of any Issue which requires a Component to be associated,  (currently this is just the 'Suspected hardware issue' Issue I believe).


spec/features/case/new_spec.rb
```diff
@@ -151,4 +151,60 @@
...
+    end
+
+    it 'does not allow selection of issues requiring a component' do
+      within('#issue') do
+        expect(
+            all('option[disabled]').map(&:text)
+        ).to have_text "#{comp_issue.name} (no components available)"
+      end
+    end
+  end
```
bobwhitelock: Lovely, thanks for adding these :raised_hands:

PR #756, opened 2018-10-08
Implement UI for setting log associations

### LINKED ISSUE #538: Ability to add multiple components to a log entry.

atoghill:
Part 2 of #538. This is a continuation of #724 and implements the association picker, previously used for cases, to set the associations for a log.

With this PR you can find the `Logs` tab in the following dashboards:
* Cluster
* Component
* Component Group
* Service

Within these tabs you will find all associated logs for that cluster. The association picker will also have the component/group/service already selected to make log creation easier:

screenshot_2018-10-08 logs - node01 - demo cluster - alces flight center

Currently this doesn't include the preview confirming what associations have been selected. However this could be added in either in this PR or a future one if it is deemed necessary.

As discussed before it may also be a good idea to separate the logs page in two. With one page displaying the logs and another for creating them.

Once merged this closes #538.

[Trello](https://trello.com/c/cdJkhtYZ/470-ui-for-creating-logs-with-multiple-associations)


app/controllers/logs_controller.rb
```diff
@@ -1,28 +1,35 @@
 require 'slack-notifier'
 class LogsController < ApplicationController
+
+  ASSOCIATION_PARAM_REGEX = /^(?<type>[A-Z][A-Za-z]*)-(?<id>[0-9]+)$/
+
   def index
-    @scope = @component || @cluster
-    @new_log = Log.new(@scope.class.name.downcase.pluralize => [@scope])
+    @scope = @component || @cluster || @component_group || @service
```
jamesremuscat: As far as I can tell, this line (and its twin in `#create`) now just set `@scope` to the same value as has already been set  by `ApplicationController#assign_scope`. Does it still have any purpose?

app/controllers/logs_controller.rb
```diff
@@ -1,28 +1,35 @@
...
+    klass = @component_group ? :component_groups : @scope.class.name.downcase.pluralize
```
jamesremuscat: `@scope.class.name.underscore.pluralize` to avoid the special-case.

Yes, `underscore` is a terrible name for that method. :man_shrugging:

atoghill:
@jamesremuscat I fiddled with adding the children component logs to the component group logs page but decided it was best to throw it into a [card](https://trello.com/c/lVU2br46/243-include-child-component-logs-on-component-group-log-page) like you suggested. The rest of your comments should hopefully be addressed now though.

app/models/component_group.rb
```diff
@@ -14,6 +14,9 @@ class ComponentGroup < ApplicationRecord
   has_many :case_associations, as: :associated_element
   has_many :cases, through: :case_associations
 
+  has_many :log_associations, as: :associated_element
+  has_many :logs, through: :log_associations
```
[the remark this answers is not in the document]
jamesremuscat: Deferred to https://trello.com/c/lVU2br46/243-include-child-component-logs-on-component-group-log-page.

app/views/logs/index.html.erb
```diff
@@ -35,8 +35,9 @@
 <% if policy(@new_log).create? %>
   <div class='card'>
     <%= render 'partials/card_header_nav', title: 'Add Log' %>
+    <div class="row">
```
jamesremuscat: You haven't closed this `<div>`.

Issue #709, opened 2018-09-27
Add "Comment and resolve" button


mjtko:
With the new case resolution workflow, it's commonplace to add a comment then resolve a case in two clicks.  This also generates two emails/slack notifications.

Ideally, we'd have a "comment and resolve" button that both added the comment and marked the case as closed with only one resultant email/slack notification.

mjtko:
I think we should bump the priority of this and deal with this sooner rather than later to reduce the quantity of unnecessary/quick-succession emails.

jamesremuscat:
Fixed by #764 and #767.

mjtko:
Reopening to remind us to split out bits that aren't fixed by #764, #767 and #790 (which I think is just the removal/replacement of multiple Slack notifications bit).

benarmston:
New ticket opened to track the only outstanding part of this issue.

PR #767, opened 2018-10-12
Allow commenting and resolving together

benarmston:
This PR adds support for commenting and resolving with one submission.  It has the same UI/UX as reopening cases. That is a dialog is shown in which the comment can be added.

I have some WIP that changes the UI/UX to add a "Comment and resolve" button next to the standard form submission button.  It isn't quite ready yet, so I've extracted what I have immediately ready here.

This PR is probably most easily reviewed after #762 is merged.

Refs #709


jamesremuscat:
If you're not the assigned engineer for a case, then you are unable to add a comment, but the resolve-case form still prompts you to enter one; if you do so, the comment gets swallowed and disappears.

We could either:
1. Hide the comment box in the resolve case form if `policy(case).comment?` is `false`
2. Allow any engineer to comment when resolving a case
3. Only allow the assigned engineer to resolve a case

For now I'm inclined to opt for 1. since it's closest to our existing behaviour.

app/views/cases/_case_state_controls.html.erb
```diff
@@ -1,16 +1,67 @@
...
+          <div class="modal-content">
+            <%= form_tag resolve_case_path(kase), method: :post do |f| %>
+              <div class="modal-header">
+                <h5 class="modal-title" id="<%= modal_label_id %>">
+                  Resolve case
+                </h5>
+                <button class="close" data-dismiss="modal">×</button>
+              </div>
+              <div class="modal-body">
+                <p>Comment</p>
```
jamesremuscat: We should make it clear that the comment is optional.

benarmston:
The above two commits fix both of the issues you have raised.  The UI/UX has changed so that there is now a "resolve" button next to the standard "submit" button.

This should make it obvious that the comment is optional, as the form can generally be submitted without a comment.

As the comment form is only shown for engineers assigned to the case, we no longer prompt for a comment that will be ignored.

This new method does expose a new potential issue.  Someone could carefully craft a request to add a comment, change some fields and resolve a case.  If the case cannot be resolved, but the request to change the fields and add a comment is legitimate, we change the fields and add the comment, but silently ignore the attempt to resolve the case.

This doesn't expose any attack vectors, and the UI only displaying the resolve button when relevant prevents this issue from being uncovered by our users.  So its probably more of an API oddity that might cause us some head scratching at some point.

---

In terms of the UI I think this is what we should be aiming for.  The UX of sending a plethora of emails still exists, but that is looking much more easily resolved now than it did before.

jamesremuscat:
This is looking really good - completely agree that this is the UI we should be aiming for, and the email UX will follow.

selection_013

I wonder if the first button should be labelled "Update and resolve case" to make it clearer that its actions are a superset of "Update case details"? @mjtko?

> Someone could carefully craft a request to add a comment, change some fields and resolve a case. If the case cannot be resolved, but the request to change the fields and add a comment is legitimate, we change the fields and add the comment, but silently ignore the attempt to resolve the case.

I don't think we need to worry about this too greatly - as you say, it's not an attack vector, and our UI prevents it from occurring in normal usage. We might consider adding a flash message if we do end up getting confused by it. (An interesting case to consider might be where updating the case attributes renders it unresolvable by e.g. unsetting `time_worked` - that might legitimately occur through the UI, and we actually handle that situation quite well with an error flash as things stand.)

benarmston:
@jamesremuscat your comment about unsetting `time_worked` when resolving has led me to realise that this PR doesn't permit updating the time worked (to a sensible value) and resolving in the same request.  I have a fix and tests for this.  Let me know if you want them including on this PR or a separate one.

jamesremuscat:
@benarmston The above is my only outstanding comment on this PR; once resolved I'll be very happy to merge it.

PR #788, opened 2018-10-15
Remove case comment policy

jamesremuscat:
The case comment policy is now unused (and the `CaseCommentsController` was erroneously still trying to call `create?` on it). This PR points CCC to the correct method on the `CasePolicy` and removes the redundant file.


benarmston:
Well spotted @jamesremuscat, I should have noticed that one myself.  It would also appear that we don't have a test to cover this.  I'll see if adding one is simple.

PR #789, opened 2018-10-15
Preserve original fields and add comment when opening related case.

### LINKED ISSUE #786: Related case replaces details rather than adds an initial comment
I don't think the UX with related cases is right yet.

I agree that the details may be different or even irrelevant (though, atm, we're preserving the subject (but #785), affected components, issue type) though I feel that , and we do have a reference to the related case, but my expectation when I open a related case is to "port" the key details from the original case (perhaps be able to update them, but not required right now) and add an initial comment.

Perhaps there's a technical reason that means we're doing it the way we are currently?

jamesremuscat:
Resolves #786.


jamesremuscat:
This solution has problems if the comment is invalid - stand by.

PR #790, opened 2018-10-16
Avoid sending a plethora of emails when updating a case

benarmston:
See comment on https://github.com/alces-software/alces-flight-center/commit/b88928495018be269efad46e0b965e9112aced97 in particular regarding the recipient logic.


benarmston:
@jamesremuscat thanks for removing the `contact_id`/`assignee_id` templates. I'd missed them.  I think we still need the comment template and tests though.  The `reopen` controller makes use of it and IIRC there was somewhere else too.

jamesremuscat:
Right, those tests definitely pass locally, and are persistently failing on Travis. Sigh. Going to prod a bit more.

benarmston:
I wonder if it could be something to do with the size of the virtual frame buffer?  I've been running tests with `xvfb-run --server-args="-screen 0 1280x1024x24" bundle exec rspec` locally.  Perhaps increasing the screen size on travis makes a difference.  Or perhaps I'm paying too much attention to the "visible" part of the error message.

jamesremuscat:
OK, I think I've finished mucking around and breaking things now :slightly_smiling_face:

PR #797, opened 2018-10-16
Fix hex-based emoji more comprehensively, insofar as they now work.

### LINKED ISSUE #776: Hex-incompatible emojis are still broken (now more badly) 😱
Something's missing and it's causing the hex-incompatible emoji still to be present for selection but to no longer render correctly. 🤔

mjtko:
Refs #759, fixes #776.


jamesremuscat:
Seems to work for me: previously the `:de:` part of the below wouldn't render.

selection_017

I'm not sure there's much we can do about the Slack notifications still using :de: though!

Just to confirm (because there was some discussion): you're not expecting any changes to be needed to the deployment process for this fix? And that in future should we ever update `preview.js.erb` we just need to also update the comment at the top of that file?

PR #798, opened 2018-10-16
Make 'related case' icons consistent

### LINKED ISSUE #784: Related case iconography inconsistency
We're using `mail-forward` for forward relations, but `history` for backward relations.  Makes more sense to me/would be more consistent to use `mail-reply`.

mjtko:


jamesremuscat:
Rebased on latest `develop` (and fixed/tidied some whitespace while I was there).

FWIW, I much prefer the `fa-history` icon to `fa-reply`, but without a corresponding `fa-future` I agree it's better to be consistent and use `fa-reply`/`fa-forward` instead.

PR #800, opened 2018-10-16
Only display related cases/pending maintenance if applicable

### LINKED ISSUE #770: Related cases item shown even if there aren't any
I think that it's a bit confusing to show the related cases field if there aren't any related cases and is unnecessarily taking up screen space.  Can we remove it under these conditions please?

mjtko:
Fixes #770.


jamesremuscat:
LGTM. Thanks!

PR #802, opened 2018-10-17
Prevent accidental case form submission by using links instead of buttons.

jamesremuscat:
As title. The buttons were accidentally triggering a submit on the case edit form; we might as well use Bootstrap's buttony links.


benarmston:
I discovered the issue with one of the reopen buttons earlier today.  Thanks for fixing it.

Issue #772, opened 2018-10-15
Changing field values has become less usable


mjtko:
On #396 I suggested "replace ALL the buttons" however, it's now a scroll-and-click operation to, for example, change the assigned engineer and feels less usable. Whoops.

I'm not sure if this is a positioning thing or if we need some/all of the buttons back and make the "update case details" button continue to, well, update ALL the details.

I guess in an ideal world we'd do something clever like have a "tick" button next to each field that only became active when a change was detected and when clicked displayed a spinner and made the change in the background.  We don't live in an ideal world, and that also causes other problems with things like "if I become the engineer then display the comment textarea now".  So that sounds complex.

The "add the buttons back" seems like a compromise that will (a) keep things more usable and (b) prevent people from submitting comments and losing the other things they've changed (primarily time worked).

benarmston:
Having spent some time trying to reorganise the positioning of the elements, I'm no longer convinced that that approach is viable.  The best that I could do is

![image](https://user-images.githubusercontent.com/287050/47012786-06acd380-d13d-11e8-95b9-df2777b98ec1.png)

which still doesn't show the submit button without scrolling.  And in certain circumstances, such as maintenance windows and related cases being present the submit buttons are even further away.

mjtko:
@jamesremuscat: yup, that sounds like the pragmatic solution for now.  And, yup, I'm keen on seeing this page get a lick of paint!

jamesremuscat:
Fixed by #803.

PR #803, opened 2018-10-17
Case editing usability

benarmston:
This PR changes the case editing UX as recently [discussed on slack](https://alces.slack.com/archives/C72GT476Y/p1539705448000100).  In brief:

1. For all fields other than "time worked" and "comment" add back a dedicated button which updates only that field.
2. Move "time worked" to the same part of the form as "comment".
3. Add a "update time worked and comment" button which updates the time worked field, if changed, and adds the comment if there is one.
4. Add a "Resolve and update time worked and add a comment" button, which updates the time worked field, if changed, adds the comment if there is one and resolves the issue.
5. The button in (4) is to be abled/disabled depending on whether there is time in the time worked field.


benarmston:
Refs #772

app/views/cases/_submit_buttons.html.erb
```diff
@@ -1,12 +1,13 @@
...
+  <% if policy(kase).resolve? && kase.open? %>
+    <%= submit_tag 'Comment and resolve',
+      name: 'restrict_commit[resolve]',
+      class: 'btn btn-secondary ml-2',
+      disabled: !kase.resolvable?
     %>
   <% end %>
-  <%= submit_tag 'Update',
-    name: 'update',
+  <%= submit_tag 'Comment',
```
jamesremuscat: This button is now shown all the time, even when we're showing a "Commenting is disabled because..." message directly above it:

selection_025

(Screenshot taken as admin; issue exists as site user too.)

Given that it's now for commenting only again, it shouldn't be shown if the user is unable to comment on the case.

app/assets/javascripts/case.js
```diff
@@ -0,0 +1,25 @@
+function initResolveButtonToggling() {
+  var hoursInput = document.querySelector('[name="case[time][hours]"]');
```
jamesremuscat: Please could we use `const` (or `let`) rather than `var`?

app/views/cases/_submit_buttons.html.erb
```diff
@@ -1,12 +1,13 @@
...
-  <% if policy(kase).resolve? && kase.resolvable? %>
-    <%= submit_tag 'Update and resolve',
-      name: 'resolve',
-      class: 'btn btn-secondary ml-2'
+<div class="text-right float-right" id="submit_buttons">
+  <% if policy(kase).resolve? && kase.open? %>
+    <%= submit_tag 'Comment and resolve',
+      name: 'restrict_commit[resolve]',
+      class: 'btn btn-secondary ml-2',
+      disabled: !kase.resolvable?
```
jamesremuscat: Disabling this button reveals a CSS bug: the `.btn-secondary:hover` overrides the `.btn-secondary:disabled` style and changes the text colour. I don't think it should (cf disabled buttons in the Reactware stuff, which don't change at all on hover).

spec/features/case/show_spec.rb
```diff
@@ -317,6 +347,82 @@
...
+
+          open_case.reload
+
+          expect(open_case.assignee).to eq(user)
+          expect(emails.count).to eq 1
+          expect(emails[0].parts.first.body.raw_source)
+            .to have_text 'Jerry has been set as the assigned engineer for this case'
+        end
+
+        it 'only changes assigned contact' do
```
jamesremuscat:
```suggestion
        it 'only changes assigned engineer' do
```

app/assets/stylesheets/cases.scss
```diff
@@ -332,3 +332,15 @@ th * {
...
+
+.table .tr-comment {
+    th, td {
+        padding-bottom: 0;
+    }
+}
+
+.table .tr-time {
+    th, td {
+        border-top: none;
```
jamesremuscat: We should also set `padding-top: 0` to reduce the still-quite-big gap between the comment box and the time fields.

app/views/cases/_submit_buttons.html.erb
benarmston: That button is for both commenting and changing the time.  So it might sometimes need to be shown when the user cannot comment (or perhaps cases are never in a state where time can be changed but comments not added).

I'll change the button so that it behaves as below:

 1. Comments are permitted: show the button with the text "Comment".
 2. Comments are not permitted, but time can be changed: show the button with the text "Update time".
 3. Neither adding comments, nor changing time is permitted: don't show the button.

app/views/cases/_submit_buttons.html.erb
benarmston: I agree that this isn't the correct behaviour. But disabled secondary buttons in Reactware change link colour on hover.  This can be seen by visiting https://staging.launch.alces-flight.com/default/launch, clicking on one of the cluster spec cards and hovering over the "Previous" button.

The bootstrap theme that we're using is at https://bootswatch.com/flatly/ which doesn't show this behaviour.  I wonder if it did in a previous version?

This probably needs fixing, but I don't think this PR, or even FC, is the right place for it.

jamesremuscat: Raised https://github.com/alces-software/flight-reactware/issues/36 to track that issue.

app/assets/stylesheets/cases.scss
benarmston: Done.

app/assets/javascripts/case.js
benarmston: Done.

app/views/cases/_submit_buttons.html.erb
benarmston: Thanks for the issue. You've saved me a job.

benarmston:
@jamesremuscat if you're happy with the new behaviour of the comment and resolve buttons, I've addressed everything in your review.

PR #804, opened 2018-10-17
Rework enter cluster checks page layout and styles.

### LINKED ISSUE #747: Better utilise screenspace on the check result submission form
A few areas of improvement:
* Merge the first two columns
* Move the `Associated Component` box below the result selection boxes
* Reduce the initial size of the text box

mjtko:
Fixes #747.

Before:

screenshot 2018-10-17 at 18 53 16

After:

screenshot 2018-10-18 at 00 48 27


app/views/clusters/enter_check_results.html.erb
```diff
@@ -123,6 +109,20 @@
                                class="error"
                                generated="true">
                         </label>
+                        <div>
```
jamesremuscat: This `div` seems redundant.

app/assets/stylesheets/cluster_checks.scss
```diff
@@ -1,3 +1,42 @@
-.result-btn.active {
-  background-color: black !important;
+.cluster-check-row {
+    td:first-child {
+        width: 35%;
+    }
+    td:nth-child(2) {
+        width: 60%;
```
jamesremuscat: Leaving only 5% for the final column means (on my browser) the "Add log entry" text gets cut off:

selection_026

Increasing it to 10% (by reducing this to 50%) also eliminates the vertical whitespace between the buttons and the `select`:

selection_027

(Worth noting that whitespace only existed because the `label` for errors was wrapping to the next line from the result buttons.)

benarmston:
@mjtko @jamesremuscat this seems as good a place to have this conversation as any.  @RuanEllis and @andrewprovis asked what the new "Add log entry" item in the associated component dropdown meant.  I didn't have an answer for them at the time.  Am I correct in thinking that it means exactly what "(Optional) Associated component" meant?

jamesremuscat:
@benarmston Yes, exactly that. Selecting an item in it will add the given comment as a log entry associated with that component.

benarmston:
Thanks @jamesremuscat I'll let them know.

Issue #748, opened 2018-10-08
Consider: Show the check result for the previous day


atoghill:
It could be useful showing the result for the previous day on each check on the check result submission form.

mjtko:
Refs #805

mjtko:
Fixed via #805.

PR #805, opened 2018-10-17
Show previous check result (with option to copy/modify value)

mjtko:
Based on #804.

Add placeholder text containing the previous check result comment (if available). Add a button to copy previous check result text into the comment box ready for modification.

screenshot 2018-10-18 at 00 50 02

Fies #748.


jamesremuscat:
Looks good - I'll hold out on merging this until #804 is merged.

PR #806, opened 2018-10-18
Change the row colour of a check to reflect selected result

### LINKED ISSUE #749: Consider: Change the colour of the check result row based on selected result
Changing the colour of the row to match the selected result could make it much easier to tell what result was selected.

atoghill:
Closes #749. When filling out the check result submission form each row will now change colour to reflect the selected result for its respective check. This looks as shown below:

screenshot_2018-10-18 enter check results - demo cluster - alces flight center

[Trello](https://trello.com/c/LX601DMW/549-change-the-colour-of-the-check-result-row-according-to-the-selected-result)


atoghill:
Of course, I'll rectify that now

atoghill:
Rebased on `develop`.

Not entirely sure I'm happy about explicitly setting the colour in this way. But this solution does work for cached results for now.

jamesremuscat:
Rebased again on `develop`.

app/views/clusters/enter_check_results.html.erb
```diff
@@ -182,6 +185,14 @@
                   elem.checked = true;
                   elem.parentElement.classList.add('active');
 
+                  if (current[fieldName] == 'Success') {
```
jamesremuscat: In general `===` rather than `==` should be used for comparison in JavaScript, because :ecosystem: - 

```javascript
1 == '1';  // true - a string can equal an integer!
1 === '1'  // false

false == 0;  // true
false === 0; // false

null == undefined;  // true
null === undefined;  // false
```

Using `==` allows the JS interpreter to use 'type coercion' which is probably not what one wants. In this particular case it's probably unimportant, but thought it would be useful to raise as a general point.

atoghill: Ah good to know, cheers

PR #807, opened 2018-10-18
Make site and cluster logos more "impactful"

### LINKED ISSUE #783: Site logo too small and causes site name to become truncated
The logo needs to be made more impactful.

I've not been able to test the cluster logo yet (#773), but assuming this is implemented in the same way, that will need addressing too.

jamesremuscat:
This PR makes some changes to how logos are used (and the overall site layout) to increase their visibility and impact. Logos are shown in more places, and are larger.

When set, site logos are shown:
- In the list of all sites
- Above the site name on the site home page
- In the title box for pages within the site scope (that aren't in a cluster scope)

When set, cluster logos are shown:
- In the list of clusters within a site on that site's home page
- Above the cluster name on the cluster home page
- In the title box for pages within the cluster scope

demo site - alces flight center - google chrome_022

demo cluster longnameyface - alces flight center - google chrome_023

alces flight center - google chrome_024

We make no assumptions about the shape of the logos - we set both a `max-width` and a `max-height`.

In making these changes, I've also changed the main application layout template to ensure the newly left-aligned title is also aligned with the main content. At lower screen sizes the "padding" to the left and right of the main content disappears, to maximise use of space.

There's much more that could be done, but I think that's best approached as part of a wider overhaul of layout and navigation in a future iteration.

Note: for the large logos, I'm not using the `withLogo` helpers; IMO this causes too much page flicker thanks to Turbolinks. This could probably be solved with time but given the above I don't think it's a priority right now. The consequence is that any logo URLs that break will result in "broken image" placeholders on a page.

Fixes #783.

Trello: https://trello.com/c/Btu0j5p5/562-embiggen-logos


jamesremuscat:
@mjtko Is the above satisfactory for this release?

PR #808, opened 2018-10-18
Cluster check hotkeys

benarmston:
This PR adds support for hotkeys when entering cluster check results.

This branch is based on the recent work done by Mark on the enter check results page, so it may be simpler to review once #804 and #805 have been merged.

The hotkeys I've added are:

 1. `alt+ctrl+c` to copy the placeholder text from the most recent previous entry.
 2. `alt+ctrl+y` to mark the check as successful and focus the next check.
 3. `alt+ctrl+n` to mark the check as failed and focus the next check.
 4. `alt+ctrl+o` to mark the check as N/A and focus the next check.

Hotkeys that I've not added are:

 1. Moving focus from the textare to the "node select list".  This can be done pressing `tab` twice.  I think this is as easy as any hot key.
 2. Focusing the next check without marking the current one as successful, failed, or n/a.  We don't permit the form being submitted without setting one of those values, so I don't think we should support navigating without setting one.
 3. Focusing a previous check result.  This is probably not common and if it is its as quick to use the mouse as it is to press a hot key N times to navigate back to the Nth previous check.
 4. Previewing the markdown comment.  I think this is a good idea though.
 5. Copy command to clip board.  Copying to the clipboard is not as simple as navigating and manipulating the DOM, so this has been left for now.

I'm happy to reconsider any of the above key bindings or to implement additional hotkeys.

The hotkeys work when any focusable item in a check result row has the focus.


jamesremuscat:
Rebased on `develop` now the precursors are all merged.

I agree with you on "hotkeys not added" 1-3, and have raised separate issues to track 4 (#812) and 5 (#811).

PR #817, opened 2018-10-19
Return false from hotkey event to ensure keypress isn't processed by browser

mjtko:
FSR ctrl+alt+<num> continues to output the number (on MacOS Chrome at least).
Returning false from the event handler ensures that a number is added to the
next textarea after it receives the focus.


atoghill:
`Ctrl` + `Alt` + `o` seems slightly broken for me also. Every other hotkey appears to work fine but the hotkey for setting a check's result to `N/A` pastes ó in the next textarea.

jamesremuscat:
For clarity (because I had to scroll up in Slack to check!): @atoghill's comment referred to behaviour _before_ this fix is applied; with the fix he experiences the expected behaviour.

Issue #816, opened 2018-10-18
Improve the copy cluster check command to clipboard feedback


benarmston:
It should look nicer.  Perhaps inline with the command? Perhaps smaller? Perhaps greyer; similar to bootstrap form help text?

Refs #811

mjtko:
Can we just flash the command element? Something like one of these?

 https://stackoverflow.com/questions/275931/how-do-you-make-an-element-flash-in-jquery

Having the "Copied!" text permanently stuck on the page doesn't feel right.

mjtko:
Fixed via #820.

PR #820, opened 2018-10-19
Change cluster check command copy notification to flash the element

benarmston:
Replace adding "Copied!" text with flashing  the command.


benarmston:
@mjtko is this a bit more like what you had in mind?

benarmston:
Refs #816

mjtko:
Yup, more like that -- I'm not sure that grey to white flash is quite noticeable enough though.

benarmston:
@mjtko does grey -> yellow -> grey look better to you?

mjtko:
Yup, LGTM!

PR #824, opened 2018-10-19
Ensure case filter dropdowns are always usable...

benarmston:
but never pretty :-(

This is the best that I'm going to get without significantly reworking things.  As there is a work around of clicking on the nav bar "view all cases" button, I think that we should leave things as they are for now and implement a better but more involved fix at some other point.

I believe that Reactstrap doesn't suffer from this problem, as it makes use of [popper.js](https://popper.js.org/) to handle its dropdowns, popovers, and tooltips.  Perhaps we should consider doing the same.


jamesremuscat:
Is this the expected new behaviour - the dropdowns (/droprights) appearing quite far away from their triggers?

cases - global - alces flight center - google chrome_031

benarmston:
Unfortunately, that's the best that I can do (without using something like popper.js to remove the dropdown from the `table-responsive` div).  It looks even worse when you go below 780px width.

I'm not convinced that this is better than the current behaviour, but I'm not the one being inconvenienced by the current behaviour.  So I'll leave it up to you and @mjtko to decide if this is worth it or not.

But I've spent enough time on this to be convinced that I'm not going to find a good easy solution.

benarmston:
Closing as the consensus is that the status quo is better than this easy fix.

PR #825, opened 2018-10-19
Fix the overflowing issue on the logs table

### LINKED ISSUE #441: Cosmetic/UX: logs table can overflow
When a "wide" entry is added to the cluster logs it can cause the table to overflow meaning that the details to the right cannot be viewed without scrolling.

Related screenshot in Slack: https://alces.slack.com/files/U025J03R1/FBUSGTSES/screen_shot_2018-07-19_at_17.01.02.png

atoghill:
The logs table should once again wrap like it did before, hopefully fixes #441 once more. 

[Trello](https://trello.com/c/w8ZsIgP7/566-stop-logs-table-from-overflowing)


mjtko:
The problem with this approach is that the gutter line numbers no longer line up.

Ideally, in this situation (and possibly other `pre` situations too, but less importantly) we'd find a way to make this do a responsive left-to-right overflow scroll thing instead of wrapping.

mjtko:
I'm actually happy to go with this solution for now, as it fixes the immediate usability problem of scrolling left and right in various places.  We should revisit it to fix it properly tho!

jamesremuscat:
As Mark says: :shipit: but I'll leave #441 open until we get a proper fix.

Issue #794, opened 2018-10-16
Paginate logs


mjtko:
As time has gone on, the number of log entries generated against clusters has increased.  We could now do with adding pagination in a similar vein to the pagination we've added for cases.

Similar to #775, have a default amount of logs that are displayed (most recent N, probably 10) but allow the user to select to view a larger number at once if they would like to.

atoghill:
The last part of this issue was resolved by #844 therefore I am closing this issue.

Issue #775, opened 2018-10-15
Add option to display all (or a large amount) of cases


mjtko:
With pagination introduced with #446 load times are much better, woo!

However, there is an argument for saying that opt-in to slow load times is sometimes useful.  For example, on retail sites, it is often possible to increase the number of items being displayed from a reasonable default to "X per page" or even "all", with the expectation that the user knows it's going to take longer to load/render the page.

I'm not sure where the default "25" number of cases came from, and this seems to reduce load/render times to ~2-3s (down from ~9s for all cases as it stands).  Reducing this to < ~1s would be even better, which would be ~10 cases afaict.  I would suggest having a means to set this to "10", "25", "50" or "all".

PR #826, opened 2018-10-19
Implement pagination for the logs page

atoghill:
This PR introduces pagination to logs as per #794. However a future branch will allow the user to select how many items they would like to display per page for both cases and logs; #775.

[Trello](https://trello.com/c/CyUP1eUX/565-paginate-the-logs-table)


app/models/log.rb
```diff
@@ -19,6 +19,8 @@ class Log < ApplicationRecord
   before_validation :assign_cluster_if_necessary
   after_initialize :assign_cluster_if_necessary
 
+  paginates_per 10
```
jamesremuscat: Is there a reason we're explicitly stating 10 here, but for cases we're using the default 25?

atoghill: As per the comments on #775 I was starting to move away from the default for (slightly) faster load times. This value is changed for cases in #844 as I wanted to focus this PR on changes to the logs page.

jamesremuscat: Thanks for the clarification :+1:

PR #836, opened 2018-10-22
Fix scrolling syntax highlighted code fences

### LINKED ISSUE #441: Cosmetic/UX: logs table can overflow

mjtko:
Fixes #441.  Refs #825.


jamesremuscat:
This leaves the case detail page looking strange when the first row contains an alert spanning the entire row:

selection_033

Everything seems crammed to the right, when there's ample space to spread things out a bit more. The headers are taking more room than they need, and the CSS you've added to specify a width is not applied because of `table-layout: fixed`.

I suspect we can overcome this by explicitly defining `col`s and defining widths on that.

mjtko:
Eww, indeed that looks unpleasant!

jamesremuscat:
Rebased on `develop` to fix conflict.

PR #844, opened 2018-10-23
Implement controls to change the number of entries displayed on a page

### LINKED ISSUE #775: Add option to display all (or a large amount) of cases

atoghill:
Based on #826 which implements pagination for the logs page.

This PR implements controls that enable the user to select, from set options, the amount of entries they would like to view per page; 10, 25, 50 or all. By default this value is now set to 10 for both cases and logs.

While slightly unconventional the selection is done via buttons rather than a select dropdown. This reduces the number of interactions required to change the value and fits slightly better with the UI in my opinion.

However there are some issues with the styling currently that could do with resolving. When there is only one page the page controls are not visible as you'd expect but the entry selection controls move over to the left hand side. The current styling should be seen as a guideline for how I thought the controls should be placed but I could not find a great solution in a timely fashion.

screenshot_2018-10-23 cases - global - alces flight center

Resolves #775, #794  when merged. 

[Trello](https://trello.com/c/4zWCoRuW/568-implement-pagination-display-controls)


jamesremuscat:
Gave a very small amount of polish to the presentation - ready to merge now.

PR #863, opened 2018-10-29
Reduce plethora of slack notifications

benarmston:
This PR reduces the number of slack notifications when a case is updated.

As the case form currently updates values one at a time, it is of limited use at present, but is likely to be much more useful once the defered changes work has been completed.

No work has been undertaken to reduce the number of emails or slack notifications when doing something with a case other than updating it.  This would be a very useful feature, but I think there is enough to review on this PR already.

Regarding the implementation, a `CaseChangesHelper` has been introduced which provides an API to query the changes made to a case.  This differs from the `ActiveModel::Dirty` API by providing an interface that 1) hides changes that are not relevant when sending emails and slack notifications; and 2) provides a consistent interface to query changes to both scalar values and change to an associated model, e.g., changes to a case's issue.

The `CaseChangesHelper` class greatly simplifies the CaseMailer class and is reused by the `SlackCaseUpdateNotes` class.

`SlackCaseUpdateNotes` is a new class which given a case, an instance of `CaseChangesHelper` and a user, will return a number of slack attachments.  The attachments describe the changes that have been made to the kase.  Between 1 and 3 attachments will be returned.  One for each of a case being resolved, commented upon, or generic changes being made. 

`SlackNotifier` has gained a new method `update_case_notification` which uses `SlackCaseUpdateNotes` to create the attachments and sends them all in a single slack notification.

Finally, a number of now unused methods have been removed from `CaseMailer` and `SlackNotifier`.

The result of this is a single slack notification, which looks very similar to the previous multiple attachments we had before.


jamesremuscat:
These changes are mostly good.

There's one example I've noticed where the notifications we're now generating are slightly less nice than before, but suspect that can be resolved through tweaking some of the parameters you've introduced.

Before when commenting and resolving a case:
selection_034

With these changes:
selection_035

I think we should:
- put the comment before the state change
- not repeat the "has been resolved...awaiting verification" line
- (if still applicable) turn the case display ID in the pretext for resolution into a link as with "New comment..."

Having additional pretext lines on the second or third attachment visually breaks up sets of changes, and I'm wondering if doing without them might make things look cleaner. What do you think? (As you say, this really only affects "comment and resolve" right now because everything else is changed one thing at a time.)

benarmston:
How about these.

Resolved with a comment:

image


Resolved without a comment:

image

Comment without resolution:

image

Perhaps resolved without a comment should have some pretext?

benarmston:
And to complete the set:

An attribute updated:

image

jamesremuscat:
Beautiful :slightly_smiling_face: 

> Perhaps resolved without a comment should have some pretext?

Perhaps... but what could it say that isn't already said by the text below it?

Happy to merge this now and we can iterate on it in the future if needed. Thanks!

PR #864, opened 2018-10-30
Chore reduce more emails and slacks

benarmston:
This PR reduces the number of emails and slack notifications sent for some case state transitions.  In particular,

 - Creating a related case now sends one email and one slack.
 - Reopening a case now sends one email and one slack.

I've also checked that closing a case and verifying a case do not send more than one email or more than one slack.

In brief:

 - Sending a creation case email has been moved from the model to the controller and lots of tests updated to accomodate.

 - The new case slack notification has been updated to include a comment attachment if a comment is provided when opening the case.

 - The case reopened email has been updated to include the comment.

 - The case reopened slack notification now uses the SlackCaseUpdateNote model to generate a single slack notification.

I've also done some refactoring on the cases controller.  I've since looked at your deferred updates work, and wonder if these might need removing.  Let me know if you want them gone.


app/views/cases/_case_state_controls.html.erb
```diff
@@ -56,7 +56,8 @@
               <p>Please explain your reasons for wanting to reopen this case:</p>
               <textarea
                 class="form-control"
-                name="comment"
+                name="case[case_comments_attributes][][text]"
+                id="case_comment_text"
```
jamesremuscat: This is what's causing the test failures - this DOM ID is the same as the one used in the normal comment form, so Capybara is finding it.

benarmston:
@jamesremuscat I'm happy to rebase and resolve conflicts after #868 is merged.

benarmston:
Going to create a new pull request to get travis to pick up correct commit.

PR #866, opened 2018-11-01
Informational cases

### LINKED ISSUE #801: Informational case workflow
When a condition is noticed on a cluster that we want to inform the customer about and/or request them to take action, we open a case.  The current workflow is to open an "other" category case and add a comment in the "Details" field.  We then "wait for a bit" to see if the customer wants/needs to respond, and resolve the case if there is no further action to be taken.

Instead, we'd like:

1. Open an "informational case" which is a tier 3 case, with a "Details" field.
2. Have the case behave as if it's immediately resolved, but with "acknowledge this notice" instead of "the issue is fixed - verify this case" and "request further information" instead of "the issue is not fixed - reopen this case".
3. Customer acknowledges, case becomes verified -or- customer requests more information, case becomes standard open tier 3.

Displaying the case state: as the case isn't really immediately "resolved", it would be preferable to display it with a state of "notice" rather than "resolved".

Handling time worked: we could set this to zero by default though it would be nice to be able to specify an amount of time worked when creating the case.

atoghill:
This PR implements the concept of `Informational` cases, used to inform the customer about anything we feel needs extra attention on a cluster.  When such a case is created it is automatically assigned as a `Tier 3` case under the new `notice` state. This behaves much like a `resolved` case but with mostly some cosmetic changes to ensure the wording of buttons still make sense in this new context.

Customers have the option to either acknowledge the notice or request further information. Verifying the case or turning it into a regular open case respectively.

Resolves #801 when merged.

#### Under the hood

The `administrative` field no longer exists and has instead been turned into `issue_type` which encapsulates the four different types of issues within Flight Center:
1. Normal
2. Administrative
3. Informational (_New_)
4. Internal

Much like with a `resolved` case after a while any case with the `notice` state will also be automatically acknowledged (verified).

#### State Badge

Uses the same colours as I used in #858 to highlight cases assigned to nobody.

screenshot_2018-11-01 demo8 informational - alces flight center 2

#### Alert

A reworded alert with the same functionality as the alert used for a `resolved` case

screenshot_2018-11-01 demo8 informational - alces flight center 1

#### To consider

There are some outstanding issues that I have left out of this PR but I feel should be considered as potential future enhancements:

- As the case is automatically set to `Tier 3`  upon creation if the cluster has the option to show chargeable info enabled the related alert and modal appear on the creation form. It might be worth changing this behaviour for admins but I don't see this causing too many problems for now.
- With the introduction of `issue_type` all issues previously returned by the `special?` method are now simply tagged as `internal` issues. Therefore it could be worth getting rid of `special?` now as mentioned in the document.
- Currently `time_worked` is set to 0 when an informational case is created. Having a way of setting this value within the creation form could be beneficial as an optional field.

#### Note

Unsure about some of the special wording for this, especially the emails and in particular the automatic verification for cases with the `notice` state.

[Trello](https://trello.com/c/ztSYQDbY/578-add-informational-cases)


app/assets/stylesheets/cases.scss
```diff
@@ -281,6 +281,11 @@
     background-color: rgba(51, 170, 51, 0.2);
     color: rgb(51, 170, 51);
   }
+
+  &.case-state--notice {
+    background-color: rgba(255, 108, 17, 0.2);
```
jamesremuscat: I'm not sure about the orange here: it might say "something is wrong" too much. The same blue as `resolved` might be better here but @mjtko might have a different opinion.

app/models/case.rb
```diff
@@ -42,12 +42,14 @@ class Case < ApplicationRecord
     state :resolved  # Has been resolved but not yet verified
     state :verified  # Customer has confirmed resolution
     state :closed  # Has been accounted for commercially, nothing more to do
+    state :notice  # To inform the customer of an event and/or to request action
```
jamesremuscat: Very minor thing but please can this be put in the correct place in the list (e.g. before `resolved`)?

app/services/case_verifier.rb
```diff
@@ -23,10 +23,15 @@ def auto_verify_cases(days_ago)
...
       upper_threshold = upper_limit&.business_days&.ago
-      Case.where(state: 'resolved').select do |kase|
-        kase.resolution_date && \
-          kase.resolution_date <= lower_threshold && \
-          (upper_threshold.nil? || kase.resolution_date > upper_threshold)
+      Case.where(state: ['resolved', 'notice']).select do |kase|
+        date_to_check = if kase.notice?
+                          kase.created_at
+                        else
+                          kase.resolution_date
```
jamesremuscat: :thinking: I think we probably ought to say that a case's resolution date is the date it was opened (e.g. `created_at`) for informational cases that haven't later been resolved again, and this logic should be within `Case#resolution_date` rather than here.

app/controllers/cases_controller.rb
```diff
@@ -105,6 +111,11 @@ def create_new
       @case.components << Component.find(component_id)
     end
 
+    if @case.informational?
+      @case.state = :notice if @case.informational?
```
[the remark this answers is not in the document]
atoghill: Yep, not sure why I forgot to remove that when I changed this from a single line conditional

app/models/case.rb
atoghill: Yeah of course, I wasn't 100% sure where to place this in the list so I dumped it at the bottom until told otherwise

db/data/20181029105305_add_informational_issue.rb
```diff
@@ -0,0 +1,22 @@
...
+  def up
+    issue = Issue.create!(
+      name: ADMIN_ISSUE_NAME,
+      requires_component: false,
+      requires_service: false,
+      service_type: nil,
+      issue_type: 'informational'
+    )
+
+    tier = issue.tiers.first
```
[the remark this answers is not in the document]
atoghill: Yep you're correct!

atoghill:
@jamesremuscat The migrations should be working as intended now. I am still working on the other changes but it should be possible to review this further now at least.

app/assets/stylesheets/cases.scss
atoghill: I'm not 100% certain having the same colour is the best idea as it reduces the glance value of the state badge. The main reason I chose the orange was so that it could be easily differentiated from the others and to indicate that attention was needed. But the intention was not to create a sense of alarm, so if @mjtko agrees that this should be adjusted then I'm sure we can find a better colour 😄

jamesremuscat: A very valid point - let's leave it as is for now unless/until we get feedback to the contrary.

app/views/cases/_verify_alert.html.erb
```diff
@@ -2,10 +2,16 @@
   <div class="alert alert-info alert-cr-info">
     <i class="fa fa-2x fa-check-circle mr-2" aria-hidden="true"></i>
     <span>
-      This case is awaiting verification.
+      This case is awaiting <%= kase.informational? ? 'acknowledgement' : 'verification' %>.
     </span>
 
-    <%= link_to 'The issue is fixed - verify this case',
+    <% link_text = if kase.issue.informational?
```
jamesremuscat: Once a customer has "request[ed] further information" on a case it should behave as a standard tier 3 case - this means using "resolve"/"verify" language from that point on. Suggest this means basing the language on the state of the case rather than its issue type (as previously).

app/views/cases/_case_state_controls.html.erb
```diff
@@ -7,7 +7,7 @@
 
 
 <% if policy(kase).verify? && kase.can_verify? %>
-  <%= link_to 'Verify this case',
+  <%= link_to kase.issue.informational? ? 'Acknowledge' : 'Verify this case',
```
jamesremuscat: Again - once reopened for further information, should use normal resolve/verify language as it's a normal tier 3 case then.

app/helpers/tabs_builder.rb
```diff
@@ -20,7 +20,7 @@ def read_only_cases
     {
       id: :cases, path: scope.scope_cases_path,
       dropdown: [
-        cases_entry('Active', %w(open resolved)),
+        cases_entry('Active', %w(open resolved notice)),
```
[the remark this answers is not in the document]
jamesremuscat: It matters to me less here, but you can if you'd like!

app/controllers/cases_controller.rb
```diff
@@ -230,6 +241,23 @@ def reopen
     end
   end
 
+  def request_info
```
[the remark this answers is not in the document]
atoghill: This is a case of me leaving something to do at the end then forgetting to actually go back and handle it

atoghill:
Rebased on `develop`. There are still changes incoming however.

mjtko:
Can we use `info-circle` for the state icon pls? I think it makes more sense than `exclamation-circle`.

app/models/case.rb
```diff
@@ -330,9 +330,13 @@ def cr_in_progress?
   end
 
   def resolution_date
-    transitions.where(event: 'resolve')
-               .order(created_at: :desc)
-               .first&.created_at
+    if notice?
```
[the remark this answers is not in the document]
jamesremuscat: My apologies - I misread `notice?`. The current behaviour is as required.

PR #868, opened 2018-11-01
Defer case updates by engineers

### LINKED ISSUE #740: Option to delay engineer responses until beginning of next business day
Sometimes we do work on cases outside of what are considered business hours by Flight Center (😱).  When work like this is performed, we'd like to be able to write a comment but not send the email notification to the customer.  Ideally, the Slack notification should still be sent at the time the work is done, even more ideally containing a note that the email notification is being deferred.

Possible implementation: check-box ("Defer response until next business day") shown outside of business hours (maybe always?), that is checked by default outside of business hours (would be unchecked during business hours if we always show it).

Responses should not all be sent at exactly 9am.  Thought: 9am + random(40) minutes.

To prevent double-replying, it should be obvious within the Flight Center UI that the comment/change is pending.  i.e. Obviously highlight comments/changes that have not yet been dispatched to the customer.

Consider: this should apply to any changes performed to a case that trigger an email notification to the customer, not just comments, though comments/resolved status notifications are the most important thing to handle.

### LINKED ISSUE #843: Undo/buffering capabilities
I can imagine that this is probably hideously complex, but let's have a think about what we can do to:

1. Allow an accidental case state transition to be undone (within a time period)
2. Not send out emails immediately to allow time for the undoing

(2) would also be handy if we were to consider coalescing emails into one that covered multiple modifications (e.g. update of subject and add a comment), or if we were to add editing of comments for a time period (a la Discourse).

jamesremuscat:
This PR introduces the concept of deferred (or "pending") case updates. Rather than taking immediate effect, changes made to a case by engineers are stored and their application is delayed.

- Changes made by engineers during business hours (Mon-Fri 0900-1700) are delayed by five minutes
- Changes made by engineers outside business hours are delayed until a randomly-chosen time between 0900 and 0945 on the next business day
- Changes made by site contacts are always applied immediately

While a case has pending updates, site users are so informed on the case detail page:
selection_037

Engineers are also displayed a top-of-page banner, and are shown details of the pending changes in the case "history" list:
selection_038

selection_039

Engineers are also given the option to apply updates immediately, or to cancel them entirely. In this way this feature acts as a sort of "undo" option if the engineer had clicked a button in error.

A case may only have one set of pending updates at a time. Engineers may not make subsequent updates before the pending updates are applied. Site contacts may make updates to the case while there are pending updates; doing so marks the pending updates as "held", and they will no longer be automatically applied. Held updates require manual intervention by engineers (since the site contact may have done something to make the pending changes no longer valid).

Slack notifications for changes are dispatched when those changes are queued.

Email notifications for changes are dispatched when those changes are applied.

Implementation note: only changes to cases that make use of the `update` controller method are deferrable. The following changes to cases are **never** deferred:

- Opening a new case
- Reopening a case
- Verifying a case
- Allocating credits to and closing a case
- Opening a related case
- Changing associated components
- Customer-driven case escalation (the `Open for comments` button)
- Enabling or disabling customer comments on a tier 2 case

Also note: pending updates are applied by a new Rake task, `alces:case_updates:apply_pending`, which is called every minute from the same cronjob that advances maintenance windows.

Resolves #740. Resolves #843.
Requirements doc: https://docs.google.com/document/d/1VLAa1ONGwqFMuOleBMKkUwMLX-6NkoIzPiP-Z82LZe4
Trello: https://trello.com/c/r6C1jzvh/577-optionally-defer-engineer-case-updates


jamesremuscat:
FSR Travis is testing this against the wrong Ruby version - it's suddenly started ignoring `.ruby-version` (but only for this branch)...

PR #872, opened 2018-11-02
Terminal service unique identifier is now unique

### LINKED ISSUE #867: Add unique constraints for TerminalService#service_type
Each terminal service config has a `service_type` attribute.  This is used by the client and API service to identify the service config and download it from Flight Center.

The intention is that the 2-tuple site/`service_type` is a unique identifier for site service configs and the 3-tuple, site/cluster/`service_type` is a unique identifier for cluster service configs.  Unfortunately, we don't currently have model or database validations to ensure that this is unique.

benarmston:
The `service_type` is used as a unique identifier.  This wasn't obvious from its name or the admin UI interface.  Its uniqueness was also not enforced.

This commit fixes the admin UI and the uniqueness issues.  The uniqueness has not been implemented at the database level as the scope depends on the value of the `type` column.

The name has not been changed as that would require coordinated releases of FC, FCC client and FCC server.  With these changes in place I don't think that we need to change the name too.

Fixes #867


jamesremuscat:
Should we also be enforcing this at the database level?

benarmston:
That would be ideal, but the scope of the uniqueness depends on the value of the `type` column.  We could create some kind of custom database check to do that, but a simple unique index won't be sufficient.

PR #883, opened 2018-11-05
Fix cluster check email threading

### LINKED ISSUE #758: Make sure Gmail doesn't thread separate clusters' daily checks emails
Gmail is grouping together the "daily checks submitted" emails into one thread.

We probably need to add a disambiguation suffix (per cluster) in the same way we do with cases.

benarmston:
It seems that gmail's grouping of email threads ignore unique prefixes,
but pays attention to unique suffixes.  Moving the cluster name from
prefix to sufix should be sufficient to achieve the threading that we
want in Gmail.

Testing that this change has the desired effect in Gmail without
deploying is awkward.  We assume that this will work and we'll revisit
if it does not.

Fixes #758


mjtko:
Can we keep the prefix too please?

Issue #875, opened 2018-11-05
Preserve comment text when failing to add when there's a pending update


mjtko:
> Your changes could not be saved as there are already pending changes for this case.

This is fair enough, but if I've not noticed (cos I'm half asleep on a Monday morning) then I could end up losing a carefully crafted comment.

Can the comment text be preserved in this case please?  Then I would have the opportunity to copy it before cancelling the existing update, then repasting.

mjtko:
Or it might be better to suppress the comment box in this situation?

benarmston:
Should we not be suppressing all of the form elements?

mjtko:
Not concerned about the others, cos the changes are not as hard to reproduce.

benarmston:
This was fixed by #890.

PR #890, opened 2018-11-05
Prevent commenting on cases with pending changes

benarmston:
If a case has pending changes, we no longer show the comment form to admin users.  This prevents the admin from carefully crafting a comment only to have it rejected due to the pending changes.

The presence of pending changes does not affect whether site contacts can add a comment.

I think `Case#allowed_to_comment?` is the correct location to add this rule. Or perhaps it should be `CasePolicy` or `CaseCommenting`?  To be honest I'm a little confused about where such business rules should live.  I'd appreciate some advice on this.


benarmston:
Fixes #875

jamesremuscat:
@benarmston I've just taken a look and pushed a commit that hopefully cleans up how we were making the "can they comment" decision (which, you're entirely correct to point out, was confusing before).

Please could you check if it makes sense to you?

benarmston:
@jamesremuscat that change makes things look much more sensible.  Thanks for explaining the separation of concerns between `CasePolicy` and `CaseCommenting`.

I can work with that separation of concerns, though I don't understand why we don't want to have the `*Policy` classes deal with the specific case (errr..., I mean deal with whether this specific case can be acted upon by this specific user).  What do we gain by doing things this way?

If you've got time to elaborate I'd like to hear it.

benarmston:
We have conflicts now. I'll rebase and fix.

jamesremuscat:
The `CaseCommenting` class (I think) predates our use of Pundit for permissions. Most of the "can-I-comment" logic was there already so the most straightforward way of tidying up was to put it all there.

It would probably make some sense to merge `CaseCommenting` into `CasePolicy` at some point, but that would be slightly more involved than the small change I made!

PR #920, opened 2018-11-09
Add ability to change log case associations

### LINKED ISSUE #840: Assign an existing log entry to a case
If a log entry has been created in advance of a case being created to cover the work performed, it would be useful to be able to assign the existing log entry to a case.

benarmston:
This allows an existing log to be added to an existing case.  Whilst not the nicest UI/UX it fixes #840.

I'll create this PR now, but I'm expecting that there will be issues with the UI/UX.  I'm currently working on a better UI/UX for this issue.


benarmston:
@jamesremuscat if you've not started reviewing this yet, I'll update the PR with what's my least unfavourite UI.

jamesremuscat:
I think this UI/UX is fine for now, and in particular, fine for the use case articulated in #840.

Couple of tweaks I think should be made before it's merged:

- We should add some text at the top along the lines of "please select all logs to be associated with this case" and a "Return to case" link (cf. case association edits) - otherwise it feels odd to land back on the list of logs after selecting one to associate. (I would have said "it should redirect back to the case page" but that would ruin trying to associate multiple logs.)

- The four full-width buttons under the list of associated components on a case takes up quite a lot of space. Perhaps we could collapse the log options (or log and maintenance options) into a split-dropdown style button? (Happy to consider alternatives.)

PR #922, opened 2018-11-09
Improve case association form tests

benarmston:
This PR adds additional tests to the case association form.  If there are any edge cases that you think should also be covered, let me know.

It also adds support for capturing screenshots for failed tests.


jamesremuscat:
Rebased on `develop`.

Issue #711, opened 2018-09-27
Roll gravatars out to other appropriate places within UI


mjtko:
Currently we show gravatars against engineer and site contact names on the case list, but don't make use of them elsewhere.

We should add them in other appropriate places in the UI for consistency and to make use of the visual cues they provide.

Also maybe add them to the HTML emails where appropriate.

jamesremuscat:
Fixed by #929 (and work leading up to it, as detailed above).

PR #929, opened 2018-11-12
Show Gravatars in emails

jamesremuscat:
This PR adds Gravatar icons to _most_ relevant places in HTML emails (per #711).

The only place I've not added them is in references to users within maintenance window emails. We generate a single piece of text for these used by both Slack and emails, and `markdown_format` it in the latter case. It would be considerable effort to try integrate the icons while maintaining a "single source of truth" for the text content of those messages (or else risk future changes leave the email and Slack notification inconsistent).

@mjtko If you feel strongly about including gravatars in those emails, I can add them to this PR.


mjtko:
Nope, FBM.

PR #934, opened 2018-11-12
[WIP] Automated sending of service plan renewal reminders

benarmston:
WIP PR opened only to run tests on travis.


app/mailers/service_plan_mailer.rb
```diff
@@ -0,0 +1,12 @@
+class ServicePlanMailer < ApplicationMailer
+  default bcc: Rails.application.config.email_bcc_address
+
+  def renewal_reminder(service_plan)
+    @service_plan = service_plan
+    @cluster = @service_plan.cluster
+    mail(
+      cc: service_plan.site.email_recipients,
```
jamesremuscat: Requirements query @mjtko: should this be to all site users (including viewers) or just primary/secondary contacts?

app/views/service_plans/_expires_soon.html.erb
```diff
@@ -0,0 +1,75 @@
...
+    <% else %>
+      The following email will be sent to the site contacts on <%=
+      @service_plan.formatted_reminder_due_at %>.
+    <% end %>
+    </p>
+    <div class="card-deck mb-2">
+      <div class="card">
+        <div class="card-body">
+          <div class="card-text reminder-email-body">
+            <%= @mail_body %>
```
jamesremuscat: When embedding the HTML email here, it leaks CSS definitions that interfere with some other parts of the site. The couple I've spotted are highlighted in the image below:
service plan 1st december 2017 to 30th november 2018 - alces flight center - google chrome_060

app/models/slack_notifier.rb
```diff
@@ -184,6 +184,27 @@ def cluster_check_submission_notification(cluster, user, text)
...
+      send_notification(
+        title: "[#{cluster.name}] Service plan expiring soon",
+        title_link: url,
+        color: '#ae8dc2',
+        text: <<~EOF.squish
+          The service plan for #{cluster.name} expires in
+          #{days_until_expiration} days (#{service_plan.formatted_end_date}).
+          A renewal reminder email will be sent to the site contacts in
+          #{days_until_notification} days
+          (#{service_plan.formatted_reminder_due_at}.  To place the on email
```
jamesremuscat:
```suggestion
          (#{service_plan.formatted_reminder_due_at}).  To place the email on
```

app/models/slack_notifier.rb
```diff
@@ -184,6 +184,27 @@ def cluster_check_submission_notification(cluster, user, text)
       )
     end
 
+    def service_plan_renewal_reminder_notification(service_plan)
+      cluster = service_plan.cluster
+      url = cluster_service_plan_url(cluster, service_plan)
+      days_until_expiration = (service_plan.end_date - Date.today).to_i
+      days_until_notification = (service_plan.reminder_due_at - Date.today).to_i
```
jamesremuscat: It's a bit of an edge case and I'm not sure we care about it, but on initial deployment we'll end up with a couple of clusters where `days_until_notification` (and possibly `days _until_expiration` too!) is negative.

lib/tasks/service_plans.rake
```diff
@@ -0,0 +1,15 @@
+namespace :alces do
+  namespace :service_plans do
+    namespace :renewals do
+      desc 'Notify admins of expiring service plans'
+      task notify_admins: :environment do
```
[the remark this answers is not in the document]
benarmston: Good spot.

app/models/slack_notifier.rb
benarmston: The cron task won't send reminders about service plans which have their end date in the past.  So whilst there would be service plans which would cause `days_until_{expiration,notification}` to be negative, the cron task won't case us to hit this code path with them.

app/views/service_plans/_expires_soon.html.erb
benarmston: How annoying :angry: I'll have a think of how to fix that.

app/models/slack_notifier.rb
jamesremuscat: I managed to make it do so!

selection_061

app/mailers/service_plan_mailer.rb
mjtko: All site users is fine.

app/models/slack_notifier.rb
jamesremuscat: I managed to make it do so!

selection_061

(Now with extra anonymisation...)

app/views/service_plan_mailer/renewal_reminder.html.erb
```diff
@@ -0,0 +1,12 @@
+<p>
+The service plan for <em><%= @cluster.name %></em> is due to expire on
```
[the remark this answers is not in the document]
mjtko: @wdm78: what do you think we should do for words here? View the plaintext version at https://github.com/alces-software/alces-flight-center/blob/f0deab577e1cec774bce9f68df48503c754563aa/app/views/service_plan_mailer/renewal_reminder.text.erb :)

app/models/slack_notifier.rb
benarmston: OK. I understand now.  I'll think on it.

app/views/service_plans/_expires_soon.html.erb
benarmston: The email is now loaded in an `iframe` to prevent the CSS cross-contamination.

PR #949, opened 2018-11-19
Make empty MOTD clearer

benarmston:
This PR changes the cluster dashboard and case form MOTD tool to make it clear when the cluster does not currently have a MOTD set.


benarmston:
I have noticed that whilst it is possible for a cluster to start with an empty MOTD it is not currently possible to remove an MOTD once set.  This seems like it is probably an oversight.  @jamesremuscat do you know if that is intentional?

jamesremuscat:
@benarmston An oversight I think - we should probably allow the empty string in the MOTD tool form.

benarmston:
Thanks @jamesremuscat I'll create a new PR for that.

PR #952, opened 2018-11-19
Fix MOTD change request updates

### LINKED ISSUE #951: It is no longer possible to mark that a MOTD change request has been completed
The button used to mark the MOTD change request as being completed uses `button_to`.  This results in a `form` tag nested inside the enclosing "edit case" form tag.  As a result clicking the button attempts to update the case, instead of marking the MOTD change request as having been applied.

The solution is to lift the `form` tag outside of the "edit case" form tag, using a modal dialog.  Similar to the way we handle other previously nested forms on that page.

benarmston:
This PR makes it possible to apply MOTD change requests once more.  I've created this branch on develop.  If you want me to rebase on master for a hotfix release, let me know.

Fixes #951


jamesremuscat:
I don't see value in a hotfix unless a customer opens an MOTD change case (which they haven't done so far)...

benarmston:
That makes sense @jamesremuscat

